PR #2, opened 2017-12-22
Add basic support for updating iOS network activity indicator

wmcginty:
Small tweaks to NetworkService to optionally allow for network activity to be reflected in the springboard activity indicator.


Hyperspace/Classes/Service/Network/NetworkActivityIndicatable.swift
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/// Manages and records the number of active network requests in relation to their effect on network indicators.
+public class NetworkActivityController: NetworkActivityObservable {
+    
+    // MARK: - Properties
+    public let delayInterval: TimeInterval
+    public private(set) var indicator: NetworkActivityIndicatable
+    
+    private let queue = DispatchQueue(label: "com.hyperspace.networkactivityindicator", qos: .userInitiated)
```
tylermilner: Wondering if we should go ahead and expand out the label a bit more (not that it matters much). We'll want to think about how to standardize this across our projects too.

`private let queue = DispatchQueue(label: "com.bottlerocketstudios.hyperspace.networkactivityindicator", qos: .userInitiated)`

Hyperspace/Classes/Service/Network/NetworkActivityIndicatable.swift
```diff
@@ -0,0 +1,84 @@
...
+//
+
+import Foundation
+
+/// Represents an object that can indicate to the user that network activity is currently taking place.
+public protocol NetworkActivityIndicatable {
+    var isNetworkActivityIndicatorVisible: Bool { get set }
+}
+
+/// Represents an object that is capable of recording the state of network requests and forwarding that state onto the corresponding indicator
```
tylermilner: Small documentation tweak ("onto" --> "to):
`/// Represents an object that is capable of recording the state of network requests and forwarding that state to the corresponding indicator`

Hyperspace/Classes/Service/Network/NetworkActivityIndicatable.swift
wmcginty: I see no reason why not.

PR #5, opened 2018-01-26
Abstract NetworkService functionality

wmcginty:
I pulled some of the conversion functionality out of NetworkService and intro a NetworkServiceStruct helper - this allows for a simpler approach to creating a delegated network stack. Because of the wide variety of implementation concerns, the framework can't really provide a way to store and execute 
'completion handlers' using the URLSessionDelegate (and sub-protocols).

Because of this, it is simplest to bypass the majority of BackendService and NetworkService's execution functions and use URLSession directly. It is still desirable to use Hyperspace's error and result handling however, and breaking these functions out of NetworkService itself allows us to do that.


Hyperspace/Classes/Service/Backend/BackendService.swift
```diff
@@ -46,7 +46,7 @@ extension BackendService: BackendServiceProtocol {
         networkService.cancelTask(for: request)
     }
     
-    private func handleResponseData<T: NetworkRequest, U>(_ data: Data, for request: T, completion: @escaping BackendServiceCompletion<U>) where T.ResponseType == U {
+    private func handleResponseData<T: NetworkRequest>(_ data: Data, for request: T, completion: @escaping BackendServiceCompletion<T.ResponseType>) {
```
tylermilner: I'm digging the simplification!

Hyperspace/Classes/Service/Network/NetworkService.swift
```diff
@@ -91,3 +68,59 @@ extension NetworkService: NetworkServiceProtocol {
...
 }
+
+/// A small helper struct which deals with the conversion of Data, URLResponse and Error objects to NetworkService successes and failures.
+public struct NetworkServiceHelper {
+    
+    /// Used to convert a client error, such as those returned by a NetworkSessionDataTask, into a NetworkServiceResult.
+    ///
+    /// - Parameter clientError: The error returned by the NetworkSessionDataTask.
+    /// - Returns: The outcome NetworkServiceResult dictated by the error.
+    public static func networkServiceFailure(for clientError: Error?) -> NetworkServiceFailure {
```
tylermilner: For consistency with `networkServiceResult(for response:)` below, would we want to return a `NetworkServiceResult` here too?

wmcginty: We could, but on the use side I ended up then having to immediately unwrap the NetworkServiceResult into the underlying NetworkServiceFailure to use it so I felt like it was a waste in this case to wrap into a Result to simply unwrap immediately after.

Maybe as an addition we could add a third function to the helper that returns the Result?

PR #11, opened 2018-02-01
Re-integrate SwiftLint

wmcginty:
Now that we've moved to a Carthage-based project setup, SwiftLint needs to be integrated slightly differently. I've chosen to rely on the users local brew installation of SwiftLint, and then to ensure we're running the latest version of SwiftLint on Travis.

This could lead to some issues where the local SwiftLint version on a PR doesn't match the version on the CI - but that should be caught before we merge.

Open to feedback on the integration.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -2,12 +2,8 @@
 line_length: 240
 
 disabled_rules:
-  - type_name
-  - trailing_whitespace
-  - force_try
-  - force_cast
-  - todo
-  - statement_position
+  - trailing_whitespace # Disables SwiftLint complaining about whitespace characters on empty lines
+  - todo # Disables auto-warning of TODO statements
 
 identifier_name:
   excluded:
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,13 @@
+# Set max line length before warning (default is 120)
+line_length: 240
+
+disabled_rules:
+  - type_name
+  - trailing_whitespace
+  - force_try
+  - force_cast
+  - todo
+  - statement_position
+
+excluded:
+  - Carthage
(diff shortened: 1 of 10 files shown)


.swiftlint.yml
```diff
@@ -0,0 +1,18 @@
+# Set max line length before warning (default is 120)
+line_length: 240
+
+disabled_rules:
```
tylermilner: Let's rework the `disabled_rules` to work on a folder basis. I don't think we should generally allow exceptions for `force_try`, `force_cast`, or `type_name`. Not sure what `statement_position` means. For reference, here's how we had the old [global .swiftlint.yml](https://github.com/BottleRocketStudios/iOS-Hyperspace/blob/fe948aebe2f1553e84a089183142df9a4c1012b9/Example/.swiftlint.yml) and the [tests .swiftlint.yml](https://github.com/BottleRocketStudios/iOS-Hyperspace/blob/fe948aebe2f1553e84a089183142df9a4c1012b9/Example/Tests/.swiftlint.yml).

Examples/Playground/Hyperspace.playground/Contents.swift
```diff
@@ -10,17 +10,27 @@ PlaygroundPage.current.needsIndefiniteExecution = true
 // MARK: - Models
 
 struct User: Decodable {
-    let id: Int
+    let identifier: Int
```
tylermilner: I'm guessing you changed this to fix a SwiftLint warning, but I'm puzzled as to why the exclusion we have for `identifier_name` in the root `.swiftlint.yml` did not allow this to go through.

wmcginty: I did originally, but thought I'd reverted them all. This must have slipped through.

Issue #6, opened 2018-01-27
Add macOS Example


tylermilner:
We've got support for iOS, tvOS, and watchOS, along with examples for all three platforms. It shouldn't be too hard to implement a macOS example if we follow the conventions of the other examples. The bulk of the work will be:
* Framework
    * Create a "Hyperspace-macOS" framework target, adding the ~12 Hyperspace source files.
    * Follow the conventions to set the target's Module name, Product name, and Info.plist location appropriately.
    * Create a "Hyperspace-macOSTests" unit test target (if it's not created by now) and add the test source files as was done in the other framework test targets.
* Example
    * Creating an interface for macOS in Interface Builder.
    * Implementing the NSViewController (or however that works on Mac).

andrew-winn-br:
Working on it. I guess we could have both Catalyst and AppKit examples...

tylermilner:
@andrew-winn-br thanks for taking this on! I'm not sure how much work you've put into it, but you might want to wait until #117 is merged (or start work off of the `feature/refactors` branch). The changes in that branch will necessitate a major version release so, if possible, I'd definitely like to include macOS/Catalyst support as part of that as well. Of course, if you're already really close to having this working, go ahead and submit a PR and we'll just make @wmcginty deal with the merge. 😈

andrew-winn-br:
@tylermilner I can hold off until 117.

tylermilner:
@andrew-winn-br #117 was merged a few days ago. Feel free to resume work on this, branching off of `master`, if you're still interested and have time.

andrew-winn-br:
@tylermilner Sounds good. Still interested.

wmcginty:
Support through SPM and Cocoapods should now exist as of #151 - would be awesome to have a macOS / Catalyst example!

andrew-winn-br:
Whoops yeah forgot about this! I can get on it.

PR #13, opened 2018-02-02
Added macOS target along with tests and examples

Adamcbrz:
#6


Examples/Hyperspace-macOSExample/ViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+        alert.messageText = title
+        alert.informativeText = message
+        alert.alertStyle = .warning
+        
+        alert.runModal()
+    }
+}
+
+extension NSApplication: NetworkActivityIndicatable {
+    public var isNetworkActivityIndicatorVisible: Bool {
```
tylermilner: We may want to consider making `isNetworkActivityIndicatorVisible` a `{ get }` only requirement on `NetworkActivityIndicatable` so that we don't have to have the no-op below.

Adamcbrz: Looks like I don't need that and it can match the other examples.

Sources/Hyperspace/Service/Network/NetworkActivityIndicatable.swift
```diff
@@ -10,7 +10,11 @@ import Foundation
 
 /// Represents an object that can indicate to the user that network activity is currently taking place.
 public protocol NetworkActivityIndicatable {
+    #if MACOS
+    //macOS does not have a activity indicator
```
wmcginty: This doesn't need to be here - the point of abstracting UIApplication's functionality into a protocol is so a custom indicator could be built on any platform.

In the case of macOS, if the client doesn't want to create a custom indicator they can just pass nil into the NetworkService.

Having said that, we will probably need to remove the default value of UIApplication.shared.

Sources/Hyperspace/Service/Network/NetworkActivityIndicatable.swift
```diff
@@ -74,11 +78,15 @@ private extension NetworkActivityController {
         delayedHide?.cancel()
         delayedHide = nil
         
+        #if MACOS
```
wmcginty: Same as above - the protocol is abstracting the use of UIApplication. The isNetworkActivityIndicatorVisible call could be forwarded to a custom indicator (as it would need to be on macOS).

Sources/Hyperspace/Service/Network/NetworkActivityIndicatable.swift
tylermilner: As for removing the default value of `UIApplication.shared`, do you think we could use the #if macros to define two different function signatures? One that sets no default for macOS and then one that sets the default for everything else?

tylermilner:
Just realized one more area that we need to update is the Readme too.

Sources/Hyperspace/Service/Network/NetworkActivityIndicatable.swift
wmcginty: We could either do that or create an extension on NetworkService that is only compiled into iOS/tvOS. The base initializer would have no default, the iOS/tvOS extension would have the default.


Names may need to be tweaked to avoid ambiguity, but I think that's a little cleaner than #if, because now that I think about it watchOS doesn't have UIApplication either.

tylermilner: Yeah, that sounds like a good approach.

tylermilner:
And...just realized we'll also need to update `.travis.yml` too. 😅

.travis.yml
```diff
@@ -37,5 +37,3 @@ script:
   - if [ $POD_LINT == "YES" ]; then
       pod lib lint;
     fi
-after_success:
```
wmcginty: I'm guessing this a product of the merge - but we'll want to ensure this is present. I think we'll also need to ensure the macOS variant is building on CI.

README.md
```diff
@@ -4,8 +4,6 @@
 [![Version](https://img.shields.io/cocoapods/v/Hyperspace.svg?style=flat)](http://cocoapods.org/pods/Hyperspace)
 [![License](https://img.shields.io/cocoapods/l/Hyperspace.svg?style=flat)](http://cocoapods.org/pods/Hyperspace)
 [![Platform](https://img.shields.io/cocoapods/p/Hyperspace.svg?style=flat)](http://cocoapods.org/pods/Hyperspace)
-[![codecov](https://codecov.io/gh/BottleRocketStudios/iOS-Hyperspace/branch/master/graph/badge.svg)](https://codecov.io/gh/BottleRocketStudios/iOS-Hyperspace)
```
wmcginty: Same again - I think it's a product of the merge. But we'll want both of these badges to remain present in the Readme

.travis.yml
wmcginty: I'm getting my branches mixed up. Ignore me.

README.md
wmcginty: I'm getting my branches mixed up. Ignore me.

.travis.yml
tylermilner: @wmcginty I guess I'm suffering from the same mixup. It looks like this is merging into 'master', which already has the codecov stuff so don't we need to get this fixed?

wmcginty: @tylermilner On third viewing, you're right. Those changes were merged into master in commit db315e so these do need to be corrected.

Adamcbrz: done

README.md
Adamcbrz: Added both codecov & codebeat from db315ed don't know why they didn't merge in properly.

wmcginty:
@tylermilner are you still looking for changes on this? It looks like the Readme changes have been fixed.

Sources/Hyperspace/Service/Backend/BackendServiceProtocol.swift
```diff
@@ -87,6 +87,6 @@ extension BackendServiceError: Equatable {
 
 extension AnyError: BackendServiceErrorInitializable {
     public init(_ backendServiceError: BackendServiceError) {
-        self.init(backendServiceError as Error)
+        self.init(backendServiceError)
```
tylermilner: @Adamcbrz Was this change intentional?

Tests/BackendServiceTests.swift
```diff
@@ -111,22 +111,4 @@ class BackendServiceTests: XCTestCase {
...
-        let rhs = BackendServiceError.dataTransformationError(NSError(domain: "Test", code: 1, userInfo: nil))
-        XCTAssertFalse(lhs == rhs)
-    }
-    
-    func test_BackendServiceProtocol_Execute() {
-        let mock = MockBackendService()
-        mock.execute(request: NetworkRequestTests.SimpleGETRequest()) { (result: Result<NetworkRequestTests.SimpleGETRequest.ResponseType, NetworkRequestTests.SimpleGETRequest.ErrorType>) in
-            XCTAssert(true)
-        }
-    }
```
tylermilner: It looks like this is causing a ~4% decrease in code coverage. Was it intentional to remove these tests?

Tests/NetworkRequestTests.swift
```diff
@@ -109,39 +84,6 @@ class NetworkRequestTests: XCTestCase {
...
-        XCTAssertNotNil(result.value)
-    }
-    
-    func test_NetworkRequest_CustomQueryEncoding() {
-        let request = CustomQueryEncodingRequest()
-        let queryString = request.encodeQueryParameterString("this is a test")
-        
-        XCTAssert(queryString == "this-is-a-test")
-    }
-
```
tylermilner: Same question for these tests.

Tests/NetworkServiceTests.swift
```diff
@@ -122,17 +122,6 @@ class NetworkServiceTests: XCTestCase {
...
-        let networkService = NetworkServiceMockSubclass(session: URLSession.shared, networkActivityIndicatable: MockNetworkActivityIndicator())
-        
-        XCTAssert(networkService.initWithSessionNetworkActivityControllerWasCalled)
-    }
-    
-    func test_NetworkServiceHelper_InvalidHTTPResponsErrorUnknownError() {
-        let networkServiceFailure = NetworkServiceHelper.networkServiceFailure(for: NSError(domain: NSURLErrorDomain, code: NSURLErrorBadURL, userInfo: nil))
-        XCTAssert(networkServiceFailure.error == .unknownError)
-    }
-
```
tylermilner: These as well.

tylermilner:
@wmcginty I noticed that there was some code removed that didn't look intentional. I assumed there were still changes to be made to correct them. Added explicit comments for them now.

tylermilner:
I think the .travis.yml still needs updating as well.

wmcginty:
Sounds good- just wanted to make sure these changes didn't fall through the cracks.

wmcginty:
Going to close this due to inactivity. We also now have macOS support after #151 .

Issue #28, opened 2018-05-08
Easily get and parse error response when receiving a failure


earlgaspard:
Whenever I receive a result with a failure from a executing a BackendService, I may have a response returned with that failure. Right now it is inconvenient to get this response from a BackendServiceError and parse it into the response model I want to use.

Example:
backendService.execute(request: request) { [weak self] (result) in
            switch result {
            case .success(let response):
                completion(response.data, nil)
            case .failure(let error):
                switch error {
                case .networkError(let networkServiceError, let httpResponse):
                    if let data = httpResponse?.data {
                        let errorResponse = try? JSONDecoder().decode(MyErrorResponse.self, from: data)
                        
                        
                    }

wmcginty:
I've been experimenting with some ways to do this as well. Currently, I think the cleanest way to do this would be to make better use of `NetworkServiceHelper`. Add a protocol on top of that object, and allow it to be injected into the `NetworkService`. This would in theory allow the user to customize how to handle responses using:

```swift 
func networkServiceResult(for response: HTTP.Response) -> NetworkServiceResult
```

Thoughts?

wmcginty:
I pushed my experiments to the  'responseHandling' branch

PR #38, opened 2018-05-14
Error handlingv2

wmcginty:
This should help address #28 .

Changes made:

- Created a new protocol, `NetworkServiceFailureInitializable` (not sold on the name). This is any error that can be initialized from a `NetworkServiceFailure` as so: ` init(networkServiceFailure:)`
- Created a new protocol, `DecodingFailureInitializable` (not sold on the name). This is any error that can be initialized from a failed Codable decode as so: `init(decodingError:data:)`.
- Removed `BackendServiceError`. It is now only implemented in tests.
- Created another associated type for `NetworkRequest`, `ErrorType`, which must be `NetworkServiceFailureInitializable`. `AnyError has been expanded to conform to this protocol, so `AnyNetworkRequest<T>` is unchanged (but lots of code has been deleted, so yay).

This combination of changes allows for a few improvements:
- If the `NetworkService` creates any sort of failure response, instead of getting a generic `BackendServiceError`, `NetworkRequest.ErrorType.init` is called with the reason for the failure. This should give easy access to the `Data` in the failure (and any other part of it).
- If the `NetworkRequest.ResponseType` conforms to `Codable` and fails to decode, `NetworkRequest.ErrorType.init` can be called with both the failure and the data attempting to be decoded. There is an extension to apply a default for when `NetworkRequest.ErrorType` conforms to `DecodingFailureInitializable`.

Oh, and it changes the interface significantly enough that, if approved, we'll want to roll it up with some of the other proposed v2 changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Enhancements
 
-* None.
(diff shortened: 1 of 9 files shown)


wmcginty:
@tylermilner I put a `CHANGELOG.md` entry together. I can look at creating a migration guide as we get closer to 2.0 release.

tylermilner:
Looks good. Maybe just merge this into a `release` branch so that we're not touching `master` until we're ready to release version 2.0?

wmcginty:
Switched over to merge back into release. Should be good to go now.

Issue #37, opened 2018-05-11
Remove Query Parameters from NetworkRequest


wmcginty:
It seems like one of the biggest complexities that NetworkRequest is currently set up to deal with is the encoding of the URL query. Would it make sense to remove that functionality from NetworkRequest and either create a separate builder object for the request's URL or leave it out of Hyperspace altogether (similarly to how URLRequest does not concern itself with the components of the URL?

It would simplify the NetworkRequest protocol drastically:

``` swift
public protocol NetworkRequest {
    
    /// The model type that this NetworkRequest will attempt to transform Data into.
    associatedtype ResponseType
    associatedtype ErrorType: Swift.Error
    
    /// The HTTP method to be use when executing this request.
    var method: HTTP.Method { get }
    
    /// The URL to use when executing this network request.
    var url: URL { get }
       
    /// The header field keys/values to use when executing this network request.
    var headers: [HTTP.HeaderKey: HTTP.HeaderValue]? { get }
    
    /// The payload body for this network request, if any.
    var body: Data? { get }
    
    /// The cache policy to use when executing this network request.
    var cachePolicy: URLRequest.CachePolicy { get }
    
    /// The timeout to use when executing this network request.
    var timeout: TimeInterval { get }
    
    /// The URLRequest that represents this network request.
    var urlRequest: URLRequest { get }
      
    /// Attempts to parse the provided Data into the associated response model type for this request.
    ///
    /// - Parameter data: The raw Data retrieved from the network.
    /// - Returns: A result indicating the successful or failed transformation of the data into the associated response type.
    func transformData(_ data: Data) -> Result<ResponseType, ErrorType>
}
```

tylermilner:
It definitely would simplify things. I like your suggestion about having a separate builder object. I wouldn't want to remove the functionality completely since I think URL-encoding query parameters is pretty standard stuff that we wouldn't want our projects to rewrite every time.

PR #40, opened 2018-05-14
RFC: Separate URL Query configuration from NetworkRequest

wmcginty:
This is an initial implementation of #37. It moves the encoding and appending of the URL query into a category on URL (maintaining the encoding strategies).

Thoughts?

### DIFF
--- a/Sources/Hyperspace/Extensions/URL+Additions.swift
+++ b/Sources/Hyperspace/Extensions/URL+Additions.swift
@@ -9,11 +9,46 @@
 import Foundation
 
 extension URL {
+    
     func appendingQueryString(_ queryString: String) -> URL {
         // Conditionally add the '?' character
         let fullQueryString = queryString.isEmpty ? "" : "?\(queryString)"
         
         guard let url = URL(string: "\(absoluteString)\(fullQueryString)") else { fatalError("Unable to create \(URL.self) from query string: \(fullQueryString)") }
         return url
+    }
+}
+
+public extension URL {
+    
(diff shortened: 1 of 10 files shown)


Sources/Hyperspace/Extensions/URL+Additions.swift
```diff
@@ -17,3 +18,37 @@ extension URL {
         return url
     }
 }
+
+public extension URL {
```
tylermilner: Now that I'm seeing this as an extension, I'm wondering if we should maybe take a hint from `Foundation` and `JSONEncoder` and create a `URLQueryEncoder` (`URLQueryParameterEncoder`?). I'm sitting here looking at `JSONEncoder` for another project and was seeing the parallels (minus needing decoding support). What are your thoughts?

wmcginty: Now that you mention it, it does make sense. It's a simplified use case of an encoder, but an encoder. My only worry in this case is it doesn't make as much sense to conform to `Encoder`. We could make a `URLQueryParameterEncoder` that is an encoder in name only?

tylermilner: Yeah, `Encoder` does seem a bit heavy for what we need. I'm fine with naming it that way, but not actually investing in making it conform to `Encoder`.

wmcginty: Made the change. Left a category on `URL` which makes it easy to use the `URLQueryParameterEncoder` to append the query.

PR #41, opened 2018-05-14
RFC: Runtime Root Key Decoding

wmcginty:
This provides an implementation for being able to being decoding a JSON blob at an arbitrary String 'rootKey'. We currently have similar functionality using the `DecodableContainerType`, but this involves the generation of a type for each possible key, and so is enforced at compile time.

This method (I have kept `DecodableContainerType` available in the interface) sacrifices some run time performance for the ability to provide that String at runtime.

While the current implementation accept only a single String parameter, the same logic could be applied to an array of Strings at the cost of additional performance.

### DIFF
--- a/Tests/DecodingTests.swift
+++ b/Tests/DecodingTests.swift
@@ -90,62 +90,78 @@ class DecodingTests: XCTestCase {
         XCTAssertNil(result.value)
     }
     
-    func test_AnyNetworkRequesWithDecodableContainer_SuccessfullyDecodesChildElement() {
+    func test_AnyNetworkRequestWithDecodableContainer_SuccessfullyDecodesChildElement() {
         let request = AnyNetworkRequest<MockObject>(method: .get, url: NetworkRequestTestDefaults.defaultURL, containerType: MockDecodableContainer.self)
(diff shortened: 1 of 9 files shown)


Tests/DecodingTests.swift
```diff
@@ -110,6 +110,120 @@ class DecodingTests: XCTestCase {
...
+        let result = request.transformData(objectJSON)
+        XCTAssertNotNil(result.value)
+    }
+    
+    func test_AnyNetworkRequesWithRootDecodableKey_FailsWithIncorrectKey() {
+        let request = AnyNetworkRequest<MockObject>(method: .get, url: NetworkRequestTestDefaults.defaultURL, rootDecodingKey: "incorrectkey")
+        let objectJSON = loadedJSONData(fromFileNamed: "RootKeyObject")
+        let result = request.transformData(objectJSON)
+        XCTAssertNil(result.value)
+        XCTAssertTrue(result.error?.error is DecodingError)
```
tylermilner: Just for a bit more of a complete test, is there any way we could check for the `valueNotFound` case? And similarly, for the test below, check for the `dataCorruptedError` case?

wmcginty:
Typos fixed and requests are now asserting on specific `DecodingError`s.

PR #43, opened 2018-05-15
Make modifying a network request simpler

wmcginty:
I'm currently looking at creating an alternate `BackendService` which will handle automatically authenticating requests. In order to do this it, it would be good if it were simpler to modify any given `NetworkRequest` object, without having to know the concrete type.

### DIFF
--- a/Sources/Hyperspace/Request/NetworkRequest.swift
+++ b/Sources/Hyperspace/Request/NetworkRequest.swift
@@ -114,12 +114,29 @@ public extension NetworkRequest {
         return request
     }
     
+    /// Adds the specified headers to the HTTP headers already attached to the `NetworkRequest`.
+    ///
+    /// - Parameter additionalHeaders: The HTTP headers to add to the request
+    /// - Returns: A new `NetworkReqest` with the combined HTTP headers. In the case of a collision, the value from `additionalHeaders` is preferred.
+    func addingHeaders(_ additionalHeaders: [HTTP.HeaderKey: HTTP.HeaderValue]) -> Self {
+        let modifiedHeaders = (headers ?? [:])?.merging(additionalHeaders) { return $1 }
+        return modifyingHeaders(modifiedHeaders)
+    }
+    
+    /// Modifies the HTTP headers on the `NetworkRequest`.
+    ///
+    /// - Parameter headers: The HTTP headers to add to the request.
+    /// - Returns: A new `NetworkReqest` with the given HTTP headers.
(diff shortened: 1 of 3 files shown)


Sources/Hyperspace/Request/NetworkRequest.swift
```diff
@@ -113,6 +113,18 @@ public extension NetworkRequest {
         
         return request
     }
+    
+    func modifyingHeaders(_ headers: [HTTP.HeaderKey: HTTP.HeaderValue]?) -> Self {
```
earlgaspard: Three things here:
1. what about adding additional headers instead of replacing all of them
2. modifyingHeaders seems like it could be renamed to either modifyHeaders or replaceHeaders
3. I assume you can pass in nil so it clears your headers and this is the same with body, so how about some method comments

wmcginty: 1) can add this.
2) this is following the API guidelines. modifyHeaders would be for operating on an inout NetworkRequest, the 'ing' gerund implies a copy will be returned.
3) You can do that, I will add comments and a changelog entry

earlgaspard: awesome

tylermilner: I kind of agree that the name `modifyingHeaders` is somewhat ambiguous. What about `usingHeaders` since the way it's setup (for this method at least) is to replace any existing headers with whatever is included in the dictionary passed in? Maybe `replacingHeaders(with headers: ...)` could be a good name too? I feel like it's hard to tell until you see it at the call site. Some experimentation below:
```
let newRequest = request.modifyingHeaders([.contentType: .applicationJSON])
```

```
let newRequest = request.usingHeaders([.contentType: .applicationJSON])
```

```
let newRequest = request.replacingHeaders(with: [.contentType: .applicationJSON])
```

Sources/Hyperspace/Request/NetworkRequest.swift
```diff
@@ -113,6 +113,18 @@ public extension NetworkRequest {
...
+        var copy = self
+        copy.headers = headers
+        return copy
+    }
+    
+    func modifyingBody(_ body: Data?) -> Self {
```
tylermilner: I'm assuming we'll want to use whatever naming pattern we decide on here too.

wmcginty: Yea. Will switch over to `usingHeaders`, `replacingHeaders(with:)` seems a little to verbose to me, and conveys the same thing.

PR #45, opened 2018-05-17
Automatic Error Recovery

wmcginty:
This is another large feature that I want to get in before 2.0 pre-release if possible. Essentially, this does the following:

- Adds a protocol `Recoverable`, which dictates when an arbitrary operation can be recovered from and re-attempted. Also adds an extension on `Recoverable` for easily preparing subsequent attempts.
- Adds an optional property to the default `BackendService` (should it be a part of the `BackendServiceProtocol`?) for a `RequestRecoveryStrategy`. This object allows for certain errors to be handled automatically by returning a specific `RecoveryDisposition` of `retry` or `fail`.

For example: If the error returned is a 401 unauthorized, the recovery strategy can add the appropriate authorization header to the request, and return `retry`. 

I can look into a subspec if we think that it belongs in one.


Sources/Hyperspace/Service/Backend/BackendService.swift
```diff
@@ -32,11 +34,29 @@ extension BackendService: BackendServiceProtocol {
...
+    }
+    
+    public func execute<T: NetworkRequest & Recoverable>(recoverable request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
+        execute(request: request) { [weak self] result in
+            switch result {
+            case .success(let response):
+                BackendServiceHelper.handleResponse(response, completion: completion)
+                
+            case .failure(let error):
+                guard let recoveryStrategy = self?.recoveryStrategy else { return completion(.failure(error)) }
```
tylermilner: For the `else` on this `guard`, do we need to pipe it through `BackendServerHelper` so that the completion block gets run on the main thread?

wmcginty: Good catch. We very much do. Will tweak that now.

PR #49, opened 2018-06-03
Converted HTTP.Status from enum to RawRepresentable struct

tylermilner:


tylermilner:
I just realized I forgot to add a changelog entry. I'll go back and do that soon.

PR #56, opened 2018-08-03
Release/2.0.0

wmcginty:
I think we're good to go here - just needs tagging and release.


Documentation/Guides/ErrorHandling.md
```diff
@@ -0,0 +1,74 @@
+### Basic Error Handling
+
+As a general rule, networking is hard. A lot can go wrong in the process of communicating between an iOS application and a backend. One of the goals of Hyperspace is to make the many different failure reasons easy to understand and easy to react to. In this section, we'll describe the steps needed to do the bare minimum to effectively handle errors.
```
earlgaspard: Hyperspace supports tvOS and watchOS also, so it doesn't make sense to say an "iOS application" here
how about dropping "iOS" - "between an application and a backend."

wmcginty: Fixed

PR #61, opened 2018-08-14
Mark Request.queryParameters as deprecated in v2

GrandLarseny:
This change will alert developers that in 2.0 and beyond they should package query parameters in the url property.

### DIFF
--- a/Sources/Hyperspace/Request/Request.swift
+++ b/Sources/Hyperspace/Request/Request.swift
@@ -36,6 +36,10 @@ public protocol Request {
     /// The model type that this Request will attempt to transform Data into.
     associatedtype ResponseType
     associatedtype ErrorType: NetworkServiceFailureInitializable
+
+    /// The query parameters for the URL. These parameters should now be specified as part of the url property.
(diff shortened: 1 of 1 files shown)


Sources/Hyperspace/Request/Request.swift
```diff
@@ -36,6 +36,10 @@ public protocol Request {
     /// The model type that this Request will attempt to transform Data into.
     associatedtype ResponseType
     associatedtype ErrorType: NetworkServiceFailureInitializable
+
+    /// The query parameters for the URL. These parameters should now be specified as part of the url property.
+    @available(*, deprecated: 2.0, message: "Query parameters should now be specified as part of the url property")
```
wmcginty: Can we format this slightly differently: 

> Query parameters should now be specified as part of the `URL` property.

GrandLarseny: Sure. My thought was to differentiate between the url property and the URL class. Maybe I can dig through and find cocoadoc syntax hilighting?

wmcginty: Oh good point. Maybe `url` ?

PR #64, opened 2018-09-05
Feature/return headers

earlgaspard:
This is a PR for returning response headers when making a network call. If this PR gets approval, I'll make updates to the change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
     [Will McGinty](https://github.com/wmcginty)
     [#47](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/47)
 
-* Add `HTTP.HeaderValue` for JSON API specification.
+* Add `HTTP.HeaderValue` for JSON API specification. 
     [Earl Gaspard](https://github.com/earlgaspard)
     [#46](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/46)
 
@@ -47,7 +47,7 @@
     [Will McGinty](https://github.com/wmcginty)
     [#51](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/52)
 
-* Add `headers` property to `HTTP.Response`.
+* Add `headers` property to `HTTP.Response`. The method signature of `Request`’s `transformData(_:)` method has changed. If you implement a custom `transformData(_:)` method, you will need to replace it with `transformData(_:serviceSuccess:)`.
     [Earl Gaspard](https://github.com/earlgaspard)
     [#64](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/64)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@
     [Will McGinty](https://github.com/wmcginty)
     [#51](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/52)
 
+* Add `headers` property to `HTTP.Response`.
+    [Earl Gaspard](https://github.com/earlgaspard)
+    [#64](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/64)
+
 ##### Bug Fixes
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
     [Will McGinty](https://github.com/wmcginty)
     [#47](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/47)
 
-* Add `HTTP.HeaderValue` for JSON API specification. 
+* Add `HTTP.HeaderValue` for JSON API specification.
     [Earl Gaspard](https://github.com/earlgaspard)
     [#46](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/46)
 
--- a/Sources/Hyperspace/HTTP/HTTP.swift
+++ b/Sources/Hyperspace/HTTP/HTTP.swift
@@ -131,7 +131,7 @@ public struct HTTP {
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
         ///   - headers: The HTTP header fields for this response.
-        public init(code: Int, data: Data?, headers: [String: String]? = nil) {
+        public init(code: Int, data: Data?, headers: [String: String]? = [:]) {
             self.code = code
             self.data = data
             self.headers = headers
--- a/Sources/Hyperspace/HTTP/HTTP.swift
+++ b/Sources/Hyperspace/HTTP/HTTP.swift
@@ -131,7 +131,7 @@ public struct HTTP {
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
         ///   - headers: The HTTP header fields for this response.
-        public init(code: Int, data: Data?, headers: [String: String]? = [:]) {
+        public init(code: Int, data: Data?, headers: [String: String]? = nil) {
             self.code = code
             self.data = data
             self.headers = headers
--- a/Sources/Hyperspace/HTTP/HTTP.swift
+++ b/Sources/Hyperspace/HTTP/HTTP.swift
@@ -105,31 +105,36 @@ public struct HTTP {
     
     /// Represents a HTTP response.
     public struct Response: Equatable {
-        
+
         /// The raw HTTP status code for this response.
         public let code: Int
-        
+
         /// The raw Data associated with the HTTP response, if any data was provided.
         public let data: Data?
-        
+
+        /// The HTTP header fields for this response.
+        public let headers: [String: String]
+
         /// The parsed HTTP status associated with this response.
         public var status: HTTP.Status {
             return HTTP.Status(code: code)
         }
-        
+
         /// A convenience property to encode the data associated with this response into a String. Can be useful for debugging.
         public var dataString: String? {
             return data.flatMap { String(data: $0, encoding: .utf8) }
         }
-        
-        /// Initialize a new HTTP.Response with any given HTTP status code and Data.
+
+        /// Initialize a new Response with any given HTTP status code and Data.
         ///
         /// - Parameters:
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
-        public init(code: Int, data: Data?) {
+        ///   - headers: The HTTP header fields for this response.
+        public init(code: Int, data: Data?, headers: [String: String]) {
             self.code = code
             self.data = data
+            self.headers = headers
         }
     }
 }
--- a/Sources/Hyperspace/HTTP/HTTP.swift
+++ b/Sources/Hyperspace/HTTP/HTTP.swift
@@ -113,7 +113,7 @@ public struct HTTP {
         public let data: Data?
 
         /// The HTTP header fields for this response.
-        public let headers: [String: String]
+        public let headers: [String: String]?
 
         /// The parsed HTTP status associated with this response.
         public var status: HTTP.Status {
@@ -131,7 +131,7 @@ public struct HTTP {
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
         ///   - headers: The HTTP header fields for this response.
-        public init(code: Int, data: Data?, headers: [String: String] = [:]) {
+        public init(code: Int, data: Data?, headers: [String: String]? = nil) {
             self.code = code
             self.data = data
             self.headers = headers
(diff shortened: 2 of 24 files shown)


Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -105,31 +105,36 @@ public struct HTTP {
...
-        /// Initialize a new HTTP.Response with any given HTTP status code and Data.
+
+        /// Initialize a new Response with any given HTTP status code and Data.
         ///
         /// - Parameters:
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
-        public init(code: Int, data: Data?) {
+        ///   - headers: The HTTP header fields for this response.
+        public init(code: Int, data: Data?, headers: [String: String]) {
```
wmcginty: In the interest of source compatibility, the headers should probably be either an optional (and defaulted to `nil`) dictionary, or at least defaulted to `[:]`.

Sources/Hyperspace/Service/Backend/BackendServiceProtocol.swift
```diff
@@ -28,7 +28,7 @@ public protocol BackendServiceErrorInitializable: Error {
 /// Represents the completion of a request executed using a BackendService.
 /// When successful, the parsed object is provided as the associated value.
 /// When request execution fails, the relevant E is provided as the associated value.
-public typealias BackendServiceCompletion<T, E: Swift.Error> = (Result<T, E>) -> Void
+public typealias BackendServiceCompletion<T, E: Swift.Error> = (Result<T, E>, HTTP.Response?) -> Void
```
wmcginty: This might be a controversial opinion, but I don't think using the `HTTP.Response` is an important enough use case that it should get returned along with the result (and have to _ ignore it frequently). I'd much rather see us implement a way where the `Response` can be used in the `transformData` closure to create the `Result<T>`.

earlgaspard: I agree that the response is going to be ignored in most cases. There are cases where I need the response along with the result so I can add the result data and a value from the response header to my custom class. I am open to finding the best way to do this. Example:
`backendService.execute(request: request) { (result, response) in
            switch result {
            case .success(let successResponse):
                let items = successResponse
                let total = response?.headers["total"]
                let customClass = CustomClass(items: items, total: total)
                completion(customClass, nil)
            case .failure(let error):
                completion(nil, error)
            }
`

wmcginty: What if instead we passed along the `NetworkServiceSuccess` object to the `transformData` ? This would allow you to encapsulate the logic of building your result as before, and would only require changes to the end user if you have written custom transform functions.

earlgaspard: I can give it a go

Tests/NetworkServiceTests.swift
```diff
@@ -32,15 +32,16 @@ class NetworkServiceTests: XCTestCase {
     
     func test_SuccessResponseWithNoData_GeneratesNoDataError() {
```
tylermilner: Test name will need to be updated.

Sources/Hyperspace/Service/Backend/BackendServiceProtocol.swift
earlgaspard: @wmcginty what would `transformData` look like if passing along `NetworkServiceSuccess`?
would the service success be passed along with `data`?

wmcginty: My initial thought was that `BackendServiceHelper` would change to something like: 

```swift
public static func handleResponseData<T: Request>(_ data: Data, networkSuccess: NetworkServiceSuccess, for request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
        let transformResult = request.transformData(data, networkSuccess) //this would need to be modified at the `Request` level as well.
        
        executeOnMain {
            switch transformResult {
            case .success(let transformedObject):
                completion(.success(transformedObject))
            case .failure(let error):
                completion(.failure(error))
            }
        }
    }

Comment (wmcginty) [reply Sources/Hyperspace/Service/Backend/BackendServiceProtocol.swift]: Side note: The only reason I would suggest sending the whole success is in case something else gets added to that object in the future (as of now its just Data and HTTP.Response)

Comment (earlgaspard) [reply Sources/Hyperspace/Service/Backend/BackendServiceProtocol.swift]: I think I got it changed

File: Sources/Hyperspace/Service/Backend/BackendServiceHelper.swift [hunk 4]
```diff
@@ -15,37 +15,37 @@ public struct BackendServiceHelper {
...
     ///
     /// - Parameters:
-    ///   - data: The raw Data retrieved from the network.
+    ///   - response: The HTTP.Response retrieved from the network.
     ///   - request: The Request that will be used to transform the Data.
     ///   - completion: The completion block to invoke when execution has finished.
-    public static func handleResponseData<T: Request>(_ data: Data, for request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
-        let transformResult = request.transformData(data)
+    public static func handleResponse<T: Request>(_ response: HTTP.Response, for request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
+        let transformResult = request.transformData(response.data ?? Data())
```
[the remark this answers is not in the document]
earlgaspard: This part has been changed where this is no longer an issue

Sources/Hyperspace/Service/Backend/BackendServiceHelper.swift
```diff
@@ -15,37 +15,37 @@ public struct BackendServiceHelper {
...
-                completion(.failure(error))
+                completion(.failure(error), response)
             }
         }
     }
     
     public static func handleResponse<T, U>(_ response: T, completion: @escaping BackendServiceCompletion<T, U>) {
         executeOnMain {
-            completion(.success(response))
+            completion(.success(response), nil)
```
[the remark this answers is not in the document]
earlgaspard: Reversed this setup so it no longer passes nil or anything

Sources/Hyperspace/Service/Network/NetworkServiceHelper.swift
```diff
@@ -29,11 +29,7 @@ public struct NetworkServiceHelper {
         case .unknown:
             return .failure(NetworkServiceFailure(error: .unknownStatusCode, response: response))
         case .success:
-            guard let data = response.data else {
-                return .failure(NetworkServiceFailure(error: .noData, response: response))
```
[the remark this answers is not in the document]
earlgaspard: Left `noData` still in there to make changes easier

Tests/NetworkServiceTests.swift
earlgaspard: The name does not need to be changed anymore now that `noData` generates an error again

earlgaspard:
@tylermilner @wmcginty please review recent changes and leave feedback

Sources/Hyperspace/Service/Backend/BackendService.swift
```diff
@@ -35,19 +35,18 @@ extension BackendService: BackendServiceProtocol {
...
+                BackendServiceHelper.handleResponseData(serviceSuccess.data, serviceSuccess: serviceSuccess, for: request, completion: completion)
             case .failure(let serviceFailure):
                 BackendServiceHelper.handleNetworkServiceFailure(serviceFailure, completion: completion)
             }
         }
     }
     
     public func execute<T: Request & Recoverable>(recoverable request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
-        execute(request: request) { [weak self] result in
+        execute(request: request) { [weak self] (result) in
```
wmcginty: Super minor, but we probably want to remove the parens to keep the style uniform.

Sources/Hyperspace/Service/Network/NetworkService.swift
```diff
@@ -49,13 +49,13 @@ extension NetworkService: NetworkServiceProtocol {
         let task = session.dataTask(with: request) { [weak self] (data, response, error) in
             self?.networkActivityController?.stop()
             
-            guard let statusCode = (response as? HTTPURLResponse)?.statusCode else {
+            guard let statusCode = (response as? HTTPURLResponse)?.statusCode, let headers = (response as? HTTPURLResponse)?.allHeaderFields as? [String: String] else {
```
wmcginty: Would it be cleaner to cast the response to `HTTPURLResponse` once, then access it twice?

Tests/RecoverableTests.swift
```diff
@@ -141,7 +142,7 @@ class RecoverableTests: XCTestCase {
         let exp = expectation(description: "backendServiceRecovery")
         backendService = BackendService(networkService: MockRecoverableNetworkService.protectedService(for: MockObject(title: "title", subtitle: "subtitle")), recoveryStrategy: MockFailureRecoveryStrategy())
         
-        backendService?.execute(recoverable: RecoverableRequest<MockObject>()) { result in
+        backendService?.execute(recoverable: RecoverableRequest<MockObject>()) { (result) in
```
wmcginty: Same as before with the ( ). I've always used without them, but we should just make it uniform.

Tests/RecoverableTests.swift
```diff
@@ -121,16 +122,16 @@ class RecoverableTests: XCTestCase {
         let subtitle = "subtitle"
         backendService = BackendService(networkService: MockRecoverableNetworkService.protectedService(for: MockObject(title: title, subtitle: subtitle)), recoveryStrategy: MockAuthorizationRecoveryStrategy())
         
-        backendService?.execute(recoverable: RecoverableRequest<MockObject>()) { result in
+        backendService?.execute(recoverable: RecoverableRequest<MockObject>()) { (result) in
```
wmcginty: Same parens.

Sources/Hyperspace/Service/Backend/BackendService.swift
earlgaspard: I forgot to change those back

Sources/Hyperspace/Service/Network/NetworkService.swift
earlgaspard: I can do that

Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -105,31 +105,36 @@ public struct HTTP {
...
-        /// Initialize a new HTTP.Response with any given HTTP status code and Data.
+
+        /// Initialize a new Response with any given HTTP status code and Data.
         ///
         /// - Parameters:
         ///   - code: The raw HTTP status code for this response.
         ///   - data: The raw Data associated with the HTTP response, if any data was provided.
-        public init(code: Int, data: Data?) {
+        ///   - headers: The HTTP header fields for this response.
+        public init(code: Int, data: Data?, headers: [String: String]? = [:]) {
```
wmcginty: This feels kind of weird. If we're going to make it optional, why not default it to nil? If we're going to use [:] as default, should we make it non-optional?

earlgaspard: It broke several tests when making the default nil but thinking about it, it should default to nil if it is optional. I'll make it nil and fix the tests.

earlgaspard:
@tylermilner have to review again?

tylermilner:
@earlgaspard yeah, I'll look over the new changes this afternoon.

Sources/Hyperspace/Request/Request.swift
```diff
@@ -65,8 +65,9 @@ public protocol Request {
     /// Attempts to parse the provided Data into the associated response model type for this request.
     ///
     /// - Parameter data: The raw Data retrieved from the network.
+    /// - Parameter serviceSuccess: The successful result of executing a Request using a NetworkService.
     /// - Returns: A result indicating the successful or failed transformation of the data into the associated response type.
-    func transformData(_ data: Data) -> Result<ResponseType, ErrorType>
+    func transformData(_ data: Data, serviceSuccess: NetworkServiceSuccess) -> Result<ResponseType, ErrorType>
```
tylermilner: I think this is a good evolution of this function for now, but we may want to revisit this if we end up doing lots of breaking changes on the road to v3.0. I could see us possibly moving/copying the `data` property on `HTTP.Response` up to `NetworkServiceSuccess` and making it non-optional. That would make `NetworkServiceSuccess` feel a bit more useful and mirror the `error` property on `NetworkServiceFailure` better. Then our transformation function could just take a `NetworkServiceSuccess`, but then we would probably want to rename the function, breaking lots of things. 😄

tylermilner: Ignore most of that. I hadn't pulled the latest changes so it looked like `NetworkServiceSuccess` didn't have the `data` property anymore. We're good!

CHANGELOG.md
```diff
@@ -47,6 +47,10 @@
     [Will McGinty](https://github.com/wmcginty)
     [#51](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/52)
 
+* Add `headers` property to `HTTP.Response`.
```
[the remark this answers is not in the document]
earlgaspard: I can add some more info

Issue #69, opened 2018-09-21
Redundant Transform Parameters


wmcginty:
After updating to Hyperspace 2.1, it feels like there is a lot of redundancy in the transform parameters.

```swift
    /// Attempts to parse the provided Data into the associated response model type for this request.
    ///
    /// - Parameter data: The raw Data retrieved from the network.
    /// - Parameter serviceSuccess: The successful result of executing a Request using a NetworkService.
    /// - Returns: A result indicating the successful or failed transformation of the data into the associated response type.
    func transformData(_ data: Data, serviceSuccess: NetworkServiceSuccess) -> Result<ResponseType, ErrorType>
```

The data is present as `Data`, `serviceSuccess.data`, and sometimes `serviceSuccess.response.data`. We should look to change the signature of the transform to:

```swift
func transformSuccess(_ serviceSuccess: NetworkServiceSuccess) -> Result<ResponseType, ErrorType>
```

This way we can unify the method name in the protocol with that of the default implementations. This also makes it much clearer when testing as you don't have to provide 'data' multiple times (where its possible each separate variable could be used independently depending on the implementation of `transform`).

Thoughts @tylermilner @earlgaspard  ?

wmcginty:
It's non-optional in `NetworkServiceSuccess`. So as long as we don't rely on `HTTP.Response` data we should be able to rely on it's presence, no?

earlgaspard:
Correct. I was thinking of HTTP.Response for some reason with it's optional data

earlgaspard:
I was attempting to change things up in NetworkServiceHelper's networkServiceResult for the case .success:. There are times you could get back no data like a DELETE call and it would be fine to do so. I don't agree that it should result in returning a NetworkServiceFailure for that case. This is where issues happened for me when changed this functionality

tylermilner:
I noticed this apparent redundancy too when I was reviewing some of the changes that went into `2.1.0`. I came to the same conclusion as @earlgaspard where there _seemed_ to be a valid reason for the way it's done now, or maybe we just weren't trying to change things up all over the place. I'll take another look.

tylermilner:
I took a stab at how this would look in #70.

wmcginty:
Damn Tyler, when I brought it up I didn't think you'd have it done a few hours later.

tylermilner:
Don't worry, @wmcginty. I left you plenty of fun unit test and documentation updates for you! 😉

[...]

PR #70, opened 2018-09-21
Removed redundant `data` parameter on Request's `transformData(_:serviceSuccess:)` method.

tylermilner:
This is some experimentation with what was discussed in #69. The unit tests and documentation still needs to be updated if we decide to implement this.


wmcginty:
This looks good to me. I like how it's basically the best of both worlds. All the 'default' implementations can still work normally and operate on `Data` (allowing them to be composed into the function if you don't want to modify that direct functionality) and you also get the ability to operate on the `HTTP.Response` if you need to.

Two thumbs up here.

Sources/Hyperspace/Request/Decoding/DecodableContainer.swift
```diff
@@ -31,9 +31,9 @@ extension AnyRequest where T: Decodable {
                                        timeout: TimeInterval = RequestDefaults.defaultTimeout,
                                        decoder: JSONDecoder = JSONDecoder(),
                                        containerType: U.Type) where U.ContainedType == T {
-        self.init(method: method, url: url, headers: headers, body: body, cachePolicy: cachePolicy, timeout: timeout) { data in
+        self.init(method: method, url: url, headers: headers, body: body, cachePolicy: cachePolicy, timeout: timeout) { success in
             do {
-                return try .success(decoder.decode(U.ContainedType.self, from: data, with: U.self))
+                return try .success(decoder.decode(U.ContainedType.self, from: success.data, with: U.self))
             } catch {
                 return .failure(AnyError(error))
```
wmcginty: Total side note, but I think now with the introduction of `DecodingFailureInitializable` this can be condensed into `Request`. I'll take a look at doing that, I have some other work I'd like to do with error handling.

tylermilner:
@wmcginty with regards to the `2.0.0` adopters, how strict do you think we should be with semantic versioning? I'm torn between if this should be another major version update or just an update to the minor version number.

wmcginty:
@tylermilner I don't think we need to be overly strict - for instance, technically 2.1 could (and did in some cases) cause compiler errors in update. Now that 2.1 is out the door, we should probably stick to the rules of thumb we've been using.

tylermilner:
@wmcginty @earlgaspard I fixed the unit tests and updated the documentation so I think this should be good to merge into `master` now. Take a look whenever you get a chance.

PR #71, opened 2018-09-21
Tweaks to `DecodingFailureInitializable`

wmcginty:
- Adds the target type information when a decoding failure occurs
- Cleans up some duplicate code to use the `RequestDefault` functionality
- Deprecates the 'seat of pants' choose a root decoding key at run time functionality.


wmcginty:
I'm not sure how to feel about the Codebeat issue. The similar code it's flagging is calling two entirely different codepaths, even if the words are similar.

Tests/DecodingFailureTests.swift
```diff
@@ -0,0 +1,46 @@
...
+        let result = transformer(objectJSON)
+        
+        guard let error = result.error else { XCTFail("The decode should fail."); return }
+        XCTAssertEqual(String(describing: error.type), String(describing: MockDecodableContainer.self))
+        XCTAssertEqual(error.data, objectJSON)
+    }
+    
+    // MARK: - Helper
+    
+    private func loadedJSONData(fromFileNamed name: String) -> Data {
```
tylermilner: It looks like we've already got this helper method in `DecodingTests.swift`. Can you go ahead and factor it out so that we don't have the same code in 2 spots?

wmcginty: Can do. Will probably add it as an extension to XCTestCase.

wmcginty:
@tylermilner Any strong opinions about the Codebeat issue? Is it against the spirit of using it to turn off the rule?

tylermilner:
@wmcginty I kind of feel like it is. I know it's caught a code duplication issue in the past that we fixed (#19) so I would prefer to leave it on, even if it flags some false positives every once in a while. I'm not sure if there's a way we can "ignore" the current issue though to allow for merging. Last I looked, it seemed like Codebeat could use some maturity in managing the rules that it uses to scan your code.

wmcginty:
@tylermilner Added a changeling entry

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Remove the type definitions deprecated in 2.0.0
 [Will McGinty](https://github.com/wmcginty)
 [#72](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/72)
+* Added failing type information to `DecodingFailureInitializable` allowing the API to make decisions based off of the type that failed to decode and deprecate dynamically keyed decoding.
```
tylermilner: Minor, but can we get a newline before this entry?

wmcginty: Fixed. Stupid merge.

PR #74, opened 2018-11-08
Integrated Danger into Travis-CI

tylermilner:
Danger-swift has hit `1.0` and is working for us on Travis-CI. This introduces a simple `Dangerfile.swift` that makes sure `CHANGELOG.md` was edited if any changes to files in the `Sources` directory are made. Over time, we can add more rules. Check out the [Danger Swift website](https://danger.systems/swift/) for more information.

I had to rework the Travis-CI configuration file a little bit to get this working properly. We're now using the [Homebrew addon](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-os-x) to install `danger-swift` and `carthage`.


Dangerfile.swift
```diff
@@ -0,0 +1,12 @@
+import Danger
+
+let danger = Danger()
+
+let allSourceFiles = danger.git.modifiedFiles + danger.git.createdFiles
+
+let changelogChanged = allSourceFiles.contains("CHANGELOG.md")
```
earlgaspard: I recommend we add comments before each rule in this file. If more rules get added then it will make telling one rule a part from another easier. It also helps understand what the rule does and why it's there.

.travis.yml
```diff
@@ -11,18 +11,26 @@ env:
     - TVOS_FRAMEWORK_SCHEME=Hyperspace_tvOS
     - WATCHOS_FRAMEWORK_SCHEME=Hyperspace_watchOS
   matrix:
-    - DESTINATION="platform=iOS Simulator,OS=11.2,name=iPhone X"                        SCHEME="$IOS_FRAMEWORK_SCHEME"      RUN_TESTS="YES" POD_LINT="YES"
+    - DESTINATION="platform=iOS Simulator,OS=11.2,name=iPhone X"                        SCHEME="$IOS_FRAMEWORK_SCHEME"      RUN_TESTS="YES" POD_LINT="YES"  RUN_DANGER="YES"
```
wmcginty: I don't know if this is just GitHub, but the spacing looks off here.

.travis.yml
```diff
@@ -11,18 +11,26 @@ env:
...
 
-    - DESTINATION="platform=watchOS Simulator,OS=4.2,name=Apple Watch Series 3 - 42mm"  SCHEME="$WATCHOS_FRAMEWORK_SCHEME"  RUN_TESTS="NO"  POD_LINT="NO"
+    - DESTINATION="platform=watchOS Simulator,OS=4.2,name=Apple Watch Series 3 - 42mm"  SCHEME="$WATCHOS_FRAMEWORK_SCHEME"  RUN_TESTS="NO"  POD_LINT="NO"   RUN_DANGER="NO"
 cache:
   directories:
     - Carthage
+    - ~/.danger-swift
+addons:
+  homebrew:
+    update: true
```
wmcginty: That seems neater than the pre-install we were doing. 👍

Dangerfile.swift
wmcginty: Might be nice to host this separately as well on the BRS GitHub so we can (eventually) move it to other projects with a consistent set up.

tylermilner: That's a good suggestion. In the future, we might even want to break things up into different methods, but I just added a comment for now.

.travis.yml
tylermilner: It's because of the long name for the Watch simulator a few lines below. It looks better when you're seeing all 3 platforms together:
image

wmcginty: Makes sense. Will merge as soon as CI passes.

Dangerfile.swift
tylermilner: @wmcginty definitely! Danger-swift has support for plugins so I think it would just be a matter of developing our own, hosting it, and then implementing a standardized Dangerfile.swift across all of our repos. The main blocker to doing that right now is just getting a BR-owned GitHub bot account that Danger can use to post comments on PRs. Right now it's using my personal bot account for testing, but I'll see if I can get the BR account created soon.

PR #81, opened 2019-01-08
add support for chained requests

pranjalsatija:
Example:
```swift
struct Person: Codable {
    let address: Address
}

struct Address: Codable {
    let street: String
}

let getPerson = PersonRequest<Person>()
let getAddress = {(person: Person) in
    return AddressRequest<Address>(person: person)
}

let street = backendService.execute(request: getPerson).onSuccess {(person) in
    return backendService.execute(request: getAddress(person))
}.onSuccess {(address) in
    return address.street
}.waitForResult()

let streetFuture = backendService.execute(request: getPerson).onSuccess {(person) in
    return backendService.execute(request: getAddress(person))
}.onSuccess {(address) in
    return address.street
}

// street is a String?
// streetFuture is a Future<String>
```


Cartfile
```diff
@@ -1 +1,2 @@
 github "antitypical/Result" ~> 4.0
+github "pranjalsatija/FutureKit" "220e79165a6d55bda03deffdd716c26e61141513"
```
wmcginty: We'll want to make sure it's either a BR Fork or the official FutureKit (which may mean we need to have a discussion on Hyperspace's deployment target.

Sources/Hyperspace/Service/Backend/BackendService.swift
```diff
@@ -31,6 +32,16 @@ public class BackendService {
 // MARK: - BackendService Conformance to BackendServiceProtocol
 
 extension BackendService: BackendServiceProtocol {
```
wmcginty: In addition, this can be an extension to `BackendServiceProtocol` as it just calls existing requirements in the `BackendServiceProtocol`. That way it doesn't need to be implemented. Whether we want to list it as a protocol requirement so it can be customized is another story.

Sources/Hyperspace/Service/Backend/BackendService.swift
```diff
@@ -31,6 +32,16 @@ public class BackendService {
...
+    public func execute<T: Request>(request: T) -> Future<T.ResponseType> {
+        let promise = Promise<T.ResponseType>()
```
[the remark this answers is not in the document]
pranjalsatija: What about this one? https://github.com/Thomvis/BrightFutures

Cartfile
pranjalsatija: BrightFutures seems to support a low enough version of iOS that we can use it without any problems.

Sources/Hyperspace/Service/Backend/BackendService.swift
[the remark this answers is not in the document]
wmcginty: BrightFutures seems to look good, I also like FutureX (same author as Nuke, which we use fairly frequently) although it's much newer.

Cartfile
wmcginty: I _think_ we're ok to push our deployment target in Hyperspace to 9.0 (possibly even 10) in 4.0, so we might be able to avoid crossing this bridge.

Sources/Hyperspace/Service/Backend/BackendService.swift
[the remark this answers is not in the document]
pranjalsatija: Replaced it with BrightFutures. The Error parameter is now parametrized.

Cartfile
pranjalsatija: BrightFutures works without any changes. I'm going to resolve this because a version update would be in a new PR anyway.

Sources/Hyperspace/Service/Backend/BackendService.swift
pranjalsatija: I moved this into an extension on `BackendServiceProtocol`. I left the function as part of the requirements for now because I don't see a reason not to, but let me know if that's an issue.

tylermilner:
Curious to hear @earlgaspard's thoughts on futures as well.

wmcginty:
@tylermilner 

I'm not sold on Future's either (mainly because of the dependency), and it may be that chaining and parallelizing requests may be something best left to the individual implementer.

I do like the idea of a subspec where X library is used to implement chaining, and implementers doing it their own way are free to submit a PR with a new subspec.

pranjalsatija:
So I think this boils down to either removing chained requests entirely, or adding support for them, but without intermediate transformations. Personally, I'd rather the second choice.

wmcginty:
The first thing I tried was something like:

```swift
struct ChainedRequest<T: Request, U: Request> {
  let initial: T
  let chainer: (T.ResponseType) -> U
}
```

This works fine for a single chain. But if you try to chain the `ChainedRequest` you start to run into problems because the 2nd request isn't actually created until the first network request executes. 

One possibility may be to make `ChainedRequest` conform to `Request`, but then you run into problems defining that conformance.

pranjalsatija:
Yeah, I also tried something like that, but it gets ugly quickly, and it’s also more verbose than Futures. I think the best we can do without a dependency is adding support for chained requests without immediate transformations.

from Pranjal Satija

wmcginty:
The other way would be to basically re-implement the basic concept of Futures into Request. Because in reality, a `Request<Response, Error>` IS just a `Future<Response, Error>` with a dependency on a BackendService to execute that work. 

I don't know what that looks like off the top of my head.

earlgaspard:
My thoughts on Futures coincide with other comments already stated. My main concerns are introducing a new dependency into a framework where we are trying to get dependent free as possible as mentioned by Tyler. There is also a learning curve dealing with Futures. So users of Hyperspace would not only have to learn how to Hyperspace but they would have to learn Futures too.

Hyperspace.podspec
```diff
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'Hyperspace'
-  s.version          = '3.0.0'
+  s.version          = '3.1.0'
```
wmcginty: One additional thing I'd say is that we normally will pool a bunch of changes together before creating a new release, so we should revert this to 3.0.0 for now.

Hyperspace.xcodeproj/project.pbxproj
```diff
@@ -644,6 +664,13 @@
 		60D6C3E6201CD77400B3B012 /* Frameworks */ = {
 			isa = PBXGroup;
 			children = (
+				CB3A954F21E5523000A70243 /* BrightFutures.framework */,
+				CB3A954D21E5522600A70243 /* BrightFutures.framework */,
+				CB3A954B21E5521B00A70243 /* BrightFutures.framework */,
+				CB0961E621E51894006CC9F5 /* FutureKit.framework */,
```
wmcginty: Looks like there are some remnants of FutureKit in the project?

pranjalsatija: Yeah, it's odd that it's there. It's not showing up in Xcode anywhere, but I'll edit the file manually.

tylermilner:
+1 for the subspec approach. We can market chained requests as being an advanced feature that's implemented using Futures and therefore not a part of the baseline Hyperspace feature set.

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## 3.1.0 (2019-01-08)
```
wmcginty: Lets move the additions to master, there are a few other changes I believe that we will want to incorporate into a release.

Tests/Helper/Mocks/MockBackendService.swift
```diff
@@ -6,6 +6,7 @@
 //  Copyright © 2018 Bottle Rocket Studios. All rights reserved.
 //
 
+import BrightFutures
```
tylermilner: Is this import necessary? If not, let's remove it.

pranjalsatija: Removed.

wmcginty:
What is going on with CodeCov? I can't seem to find a way to force it to re-run its analysis.

tylermilner:
Maybe if we push another commit, it will re-trigger the codebeat analysis?

wmcginty:
That's the best answer I can come up with. Quick, somebody find a typo!

pranjalsatija:
@tylermilner @wmcginty 🎉

PR #84, opened 2019-02-01
SSL certificate pinning

wmcginty:
Looking to solicit feedback as I continue to build this feature out.


Tests/CertificateHashTests.swift
```diff
@@ -0,0 +1,82 @@
...
+            let result = SecTrustCreateWithCertificates(certificates as CFTypeRef, SecPolicyCreateSSL(true, nil), &trust)
+            
+            guard result == errSecSuccess else { return nil }
+            
+            if let trust = trust {
+                if SecTrustSetAnchorCertificates(trust, anchorCertificates as CFArray) != errSecSuccess {
+                    return nil
+                }
+                
+                let verifyTime: CFAbsoluteTime = 475163640.0
```
tylermilner: Might be nice to add a comment for the date here. I'm assuming we'll eventually need to update it when the current `der` certs expire and we have to drop in new ones.

Tests/PinningTests.swift
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    func test_PinningConfiguration_respectsSetSemantics() {
+        let config = PinningConfiguration(domainConfigurations: [.init(domain: defaultHost, pinningHashes: [Data()]),
+                                                                 .init(domain: defaultHost, pinningHashes: [Data(bytes: [1, 2, 3, 4])])])
+        XCTAssertEqual(config.domainConfigurations.count, 1)
+        XCTAssertEqual(config.domainConfigurations.first, .init(domain: defaultHost, pinningHashes: [Data()]))
+    }
+
+    func test_PinningConfiguration_findsCorrectDomainConfigurationI() {
```
tylermilner: Potentially incomplete method name (extra `I` at the end).

Tests/PinningTests.swift
```diff
@@ -0,0 +1,155 @@
...
+        let config2 = PinningConfiguration(domainConfigurations: [domainConfig2])
+
+        XCTAssertTrue(config2.shouldValidateCertificate(forHost: defaultHost, at: Date()))
+        XCTAssertTrue(config2.shouldValidateCertificate(forHost: defaultHost, at: Date().addingTimeInterval(100)))
+        XCTAssertFalse(config2.shouldValidateCertificate(forHost: secondaryHost, at: Date()))
+    }
+    
+    func test_PinningConfiguration_properlyValidatesCertificates() {
+        guard let googleCert = certificate(named: "google"), let appleCert = certificate(named: "apple"),
+            let domainConfig = try? PinningConfiguration.DomainConfiguration(domain: defaultHost, certificates: [googleCert])else { return XCTFail("Unable to load testing certificate") }
```
tylermilner: Missing space before `else`.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -93,3 +94,32 @@ extension UIViewController {
 }
 
 extension UIApplication: NetworkActivityIndicatable { /* No extra conformance needed. */ }
+
+extension ViewController: URLSessionDelegate {
+    
+    func urlSession(_ session: URLSession, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
+        guard let cert = certificate(named: "cloudflare") else { return completionHandler(.performDefaultHandling, nil) }
```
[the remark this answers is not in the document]
wmcginty: For sure. I'm still working through the example (and if we can integrate straight into the NetworkService). I'll keep that in mind as I rework this.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -93,3 +94,32 @@ extension UIViewController {
...
+        do {
+            let configuration = try PinningConfiguration(domainConfigurations: [.init(domain: "jsonplaceholder.typicode.com", enforced: false, certificates: [cert])])
+            let validator = CertificateValidator(configuration: configuration)
+            if validator.handle(challenge: challenge, handler: completionHandler) {
+                print("handled")
+            } else {
+                completionHandler(.performDefaultHandling, nil)
```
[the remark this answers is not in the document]
wmcginty: This is if the authentication challenge isn't a server trust, so we do nothing.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -93,3 +94,32 @@ extension UIViewController {
...
+            }
+        } catch {
+            completionHandler(.performDefaultHandling, nil)
```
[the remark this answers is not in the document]
wmcginty: Agreed.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -93,3 +94,32 @@ extension UIViewController {
 }
 
 extension UIApplication: NetworkActivityIndicatable { /* No extra conformance needed. */ }
+
+extension ViewController: URLSessionDelegate {
+    
+    func urlSession(_ session: URLSession, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
+        guard let cert = certificate(named: "cloudflare") else { return completionHandler(.performDefaultHandling, nil) }
+        do {
+            let configuration = try PinningConfiguration(domainConfigurations: [.init(domain: "jsonplaceholder.typicode.com", enforced: false, certificates: [cert])])
```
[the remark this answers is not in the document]
wmcginty: Yea, I'm thinking that if you don't provide a URLSession into the NetworkService we can create one with a delegate which can handle the pinning. Then you'd just need to provide the hashes with which to pin.

Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift
```diff
@@ -0,0 +1,141 @@
...
+        /// Determines whether any failures to pin (`.block` as a `ValidationDecision`) will cause the connection to be blocked. Setting this value to 'false' will allow connections that fail pinning validation. Defaults to `true`.
+        let enforced: Bool
+        
+        /// Any subdomains of the configuration's domain should be included. For example, dev.apple.com would be included for a configuration for apple.com. Defaults to `true`.
+        let includeSubdomains: Bool
+        
+        /// The public key hashes for which to match the presented remote certificate.
+        let pinningHashes: [Data]
+        
+        /// The date of expiration for the certificate in question. For any dates after the expiration, pinning will not be attempted, and the connection allowed.
```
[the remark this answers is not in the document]
wmcginty: The idea was that it won't break your app if you left the certificates expire. Open to suggestions.

Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift
```diff
@@ -0,0 +1,141 @@
...
+            return .performDefaultHandling
+        }
+        
+        // MARK: Hashable
+        public var hashValue: Int {
+            return domain.hashValue
+        }
+        
+        public static func == (lhs: DomainConfiguration, rhs: DomainConfiguration) -> Bool {
+            return lhs.domain == rhs.domain
```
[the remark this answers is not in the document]
wmcginty: Currently we do an Array -> Set conversion when you initialize a configuration so we don't end up with multiple configs for a single domain. The default equatable conformance doesn't catch them as equal unless we're only checking domain. I can probably do a manual unique check instead.

Tests/CertificateHashTests.swift
wmcginty: Yup, will do.

wmcginty:
This is moving right along. I think everything but the new `TrustValidatingNetworkService` has been tested and I updated the example as well. @tylermilner @rmirabelli let me know your thoughts!

Examples/Hyperspace-iOSExample/ViewController.swift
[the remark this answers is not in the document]
tylermilner: That makes sense. To make it a little more clear, we could potentially break up the logic in `validator.handle(challenge: challenge, handler: completionHandler` into 2 parts - one to check if we should handle it and another to actually handle it. I'm always a little weary of function calls inside of `if` statements that actually perform some work in addition to returning a `true`/`false` value.

I might just be nitpicking at this point though so I'll leave it to you to decide if we should change it.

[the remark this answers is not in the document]
wmcginty: I think that makes sense

Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift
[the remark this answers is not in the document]
tylermilner: Looking at the [code for TrustKit](https://github.com/datatheorem/TrustKit/blob/master/TrustKit/TSKPinningValidator.m#L115), I think they also don't attempt pinning if the cert is expired. Haven't verified in person though.

Maybe it's behavior that we should allow configuration for? Some sort of `CertificateExpirationPolicy` enum or something? I'm also curious as to which approach is considered a best practice in the industry. As far as I can tell, there is no standard, but rather just a tradeoff between security and convenience that you have to consciously make...

Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift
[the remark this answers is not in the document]
tylermilner: Ah, so it's the `domainConfigurations` property that doesn't provide a good default `Equatable` implementation? That's unfortunate since it seems like the whole idea behind Sets is to make that stuff easier.

Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift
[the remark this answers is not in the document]
wmcginty: It gets a little complicated by the fact that we can't retrieve the expiration from the certificate. We could do something like:

```enum CertificateExpirationPolicy {
     case block
     case allow(after date: Date)
} ```

File: Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift (hunk 9 above)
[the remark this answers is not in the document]
Comment (wmcginty) [reply Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift]: Well, I guess usually it'd be fine - it was just with the default conformance if I had two different `DomainConfiguration` objects with the same `domain` property, they wouldn't be seen as equal and so the Set wouldn't do the filtering for me. 

It feels like enough of a 'why would you do this?' that I've gotten rid of the change and put an assertion and a comment in the headers instead.

File: Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift (hunk 8 above)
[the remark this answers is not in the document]
Comment (tylermilner) [reply Sources/Hyperspace/Service/Certificate Pinning/PinningConfiguration.swift]: I'm having trouble understanding why we'd need the `Date` associated value for the `allow` case. Don't we already have that information in the `expiration` property itself?

Or are you saying use that enum as a replacement for the `expiration` property where we'd bundle the policy and date together? Something like:
```swift
enum CertificateExpirationPolicy {
    case block(after date: Date)
    case allow(after date: Date)
}
```
```swift
public struct PinningConfiguration {
    public struct DomainConfiguration: Hashable {
        let domain: String
        let enforced: Bool
        let includeSubdomains: Bool
        let pinningHashes: [Data]
        let expiration: CertificateExpirationPolicy?

        // ...
    }

    // ...
}
```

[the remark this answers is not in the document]
wmcginty: Yup, basically that.

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+//  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+
+@available(iOSApplicationExtension 10.0, *)
+@available(tvOSApplicationExtension 10.0, *)
+@available(watchOSApplicationExtension 3.0, *)
+
+/* This class performs validation on authentication challenges prestented to it. In addition to ensuring that the challenge is trusted by the operating system, it will
```
tylermilner: Typo: "prestented" --> "presented".

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+    ensure that the certificate being presented as part of the SSL/TLS authentication challenge is recognized by the device. */
+public class TrustValidator {
+    
+    // MARK: - ValidationDecision Subtype
+    
+    public enum Decision {
+
```
tylermilner: Should we remove this whitespace for consistency with how other enums are defined?

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+        case allow(URLCredential) /// The certificate has passed validation, and the authentication challge should be allowed with the given credentials
```
tylermilner: Typo: "challge" --> "challenge".

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+    
+    /// Determines if the validator can handle a given `AuthenticationChallenge`.
+    ///
+    /// - Parameter challenge: The `URLAuthenticationChallenge` presented to the `URLSession` object with which this validator is associated.
+    /// - Returns: Returns `true` when the 'server trust' authentication challenge has been handled. Returns `false` for all other types of authentication challenge.
+    public func canHandle(challenge: AuthenticationChallenge) -> Bool {
+        return challenge.isServerTrustAuthentication && challenge.serverTrust != nil
+    }
+    
+    /// Allows the `TrustValidator` the chance to validate a given `URLAuthenticationChallenge` against it's local certificate.
```
tylermilner: Minor typo: "it's" --> "its".

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// MARK: - SecTrust Utility
+
+extension SecTrust {
+    
+    /// Evaluates `self` and returns `true` if the evaluation succeeds with a value of `.unspecified` or `.proceed`.
+    var isValid: Bool {
+        var result = SecTrustResultType.invalid
+        let status = SecTrustEvaluate(self, &result)
```
tylermilner: Potential [code duplication](https://github.com/BottleRocketStudios/iOS-Hyperspace/pull/84/files#diff-ec0b80b5158fa4e1235d54e3968fccb4R112) with the `CertificateHasher`? See comment above about the result of `SecTrustEvaulate(...)` not being used.

Tests/Helper/Trusts.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  XCTestCase+Certs.swift
```
tylermilner: Comment-file name mismatch.

Tests/Helper/Trusts.swift
```diff
@@ -0,0 +1,78 @@
...
+        let policy = SecPolicyCreateBasicX509()
+        
+        var trust: SecTrust?
+        SecTrustCreateWithCertificates(certificates as CFTypeRef, policy, &trust)
+        
+        SecTrustSetAnchorCertificates(trust!, [TestCertificates.root] as CFArray)
+        SecTrustSetAnchorCertificatesOnly(trust!, true)
+        
+        /// All of our leaf, intermediate and root certificates are valid on `verifyDate`. Hardcode the verification time so that we don't have to continually replace certificates as they expire.
+        let verifyTime: CFAbsoluteTime = 570998974.425934
```
tylermilner: It might be nice to have an additional comment explaining what date this epoch timestamp represents.

Tests/PinningTests.swift
```diff
@@ -0,0 +1,252 @@
...
+        XCTAssertTrue(domainConfig.validate(against: TestCertificates.google))
+        XCTAssertFalse(domainConfig.validate(against: TestCertificates.apple))
+    }
+    
+    func test_TrustConfiguration_properlyValidatesCertificateHashes() {
+        let domainConfig = TrustConfiguration.DomainConfiguration(domain: defaultHost, encodedPinningHashes: ["ivJZzhltgbIeXZGekPcWiLySsZ846YXSsGgyL9bjqEY="])
+        XCTAssertTrue(domainConfig.validate(against: TestCertificates.google))
+        XCTAssertFalse(domainConfig.validate(against: TestCertificates.apple))
+    }
+
```
tylermilner: I see the above 2 tests test the `TrustConfiguration.DomainConfiguration` initializers. Do we also need one more test to test the remaining initializer that takes `pinningHashes`?
```swift
init(domain: String, enforced: Bool = true, includeSubdomains: Bool = true, pinningHashes: [Data], expirationPolicy: CertificateExpirationPolicy = .block)
```

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+            self.enforced = enforced
+            self.includeSubdomains = includeSubdomains
+            self.pinningHashes = pinningHashes
+            self.expirationPolicy = expirationPolicy
+        }
+        
+        // MARK: - Interface
+        
+        func shouldValidateCertificate(forHost host: String) -> Bool {
+            guard includeSubdomains else { return domain == host }
```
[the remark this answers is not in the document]
wmcginty: I think for now a comment should suffice, because the host always comes in with a particular format in the `URLAuthenticationChallenge`.

Tests/PinningTests.swift
wmcginty: I don't believe so as the `encodedPinningHashes` initializer runs through the `pinningHashes` initializer, just with the added benefit of it being easier to create input.

Sources/Certificate Pinning/CertificateHasher.swift
```diff
@@ -0,0 +1,123 @@
...
+        
+        return publicKeyHash as Data
+    }
+    
+    static func secKey(from certificate: SecCertificate) -> SecKey? {
+        var trust: SecTrust?
+        let status = SecTrustCreateWithCertificates(certificate, SecPolicyCreateBasicX509(), &trust)
+        
+        guard let securityTrust = trust, status == errSecSuccess else { return nil }
+        var result = SecTrustResultType.invalid
```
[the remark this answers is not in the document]
wmcginty: Looks like this has exposed some trust validation issues in the tests, will take a look.

wmcginty:
Nevermind @tylermilner. We were not evaluating there in some tests (we still need to create pinning hashes for untrusted/expired certificates). The explicit call to validation is made in `TrustValidator.evaluate(_:forHost:date:)`, line 73 to ensure that any `SecTrust` objects as part of authentication challenges are trusted by the OS, but this does not affect our ability to test.

Tests/Helper/Trusts.swift
wmcginty: Done

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -14,10 +14,27 @@ class ViewController: UIViewController {
     // MARK: - IBOutlets
     
     @IBOutlet private var postTextField: UITextField!
+    @IBOutlet private var serverTrustValidationToggle: UISwitch!
     
     // MARK: - Properties
     
     private let backendService = BackendService(networkService: NetworkService(networkActivityIndicatable: UIApplication.shared))
+    private var trustValidatingBackendService: BackendService?
```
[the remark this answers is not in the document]
wmcginty: Not sure why I didn't do this in the first place - updated.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -19,10 +19,16 @@ class ViewController: UIViewController {
     // MARK: - Properties
     
     private let backendService = BackendService(networkService: NetworkService(networkActivityIndicatable: UIApplication.shared))
-    private var trustValidatingBackendService: BackendService?
+    private lazy var trustValidatingBackendService: BackendService = {
+        // Create's a trust configuration whereby requests to the domain 'jsonplaceholder.typicode.com' will be validated to ensure they are being served a certificate matchin 'jsonplaceholder.der'
```
tylermilner: Small typo in comment: "matchin" --> "match in".

wmcginty: Typo fixed.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -14,10 +14,33 @@ class ViewController: UIViewController {
     // MARK: - IBOutlets
     
     @IBOutlet private var postTextField: UITextField!
+    @IBOutlet private var serverTrustValidationToggle: UISwitch!
     
     // MARK: - Properties
     
     private let backendService = BackendService(networkService: NetworkService(networkActivityIndicatable: UIApplication.shared))
+    private lazy var trustValidatingBackendService: BackendService = {
+        // Create's a trust configuration whereby requests to the domain 'jsonplaceholder.typicode.com' will be validated to ensure they are being served a certificate matching 'jsonplaceholder.der'
```
earlgaspard: Typo: Create's -> Creates

wmcginty: Also fixed

wmcginty:
@earlgaspard Let me know if you still want to take a look and I'll hold off merging.

earlgaspard:
@wmcginty I still want to give it a look.

Examples/Hyperspace-iOSExample/ViewController.swift
```diff
@@ -14,10 +14,33 @@ class ViewController: UIViewController {
...
+    
+    var preferredBackendService: BackendService {
+        return serverTrustValidationToggle.isOn ? trustValidatingBackendService : backendService
+    }
+    
+    // MARK: - Lifecycle
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        //Create's a trust configuration whereby requests to the domain 'jsonplaceholder.typicode.com' will be validated to ensure they are being served a certificate matchin 'jsonplaceholder.der'
```
earlgaspard: The code here is a duplicate of the setup in the lazy property for `trustValidatingBackendService`

Sources/Certificate Pinning/AuthenticationChallenge.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Created by Will McGinty on 1/31/19.
+//  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+
+/// This protocol represents a generic authentication challenge, such as those presented as part of the SSL and TLS handshakes.
+public protocol AuthenticationChallenge {
+    
+    /// The host of the remote connection
```
earlgaspard: Can add a period to be consistent with other comments here.
```
/// The host of the remote connection.
```

Sources/Certificate Pinning/CertificateHasher.swift
```diff
@@ -0,0 +1,129 @@
...
+//  Hyperspace
+//
+//  Created by Will McGinty on 1/30/19.
+//  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+import CommonCrypto
+
+@available(iOSApplicationExtension 10.0, *)
```
earlgaspard: what do you think about combining these into one line?
```
@available(iOSApplicationExtension 10.0, tvOSApplicationExtension 10.0, watchOSApplicationExtension 3.0, *)
```

Sources/Certificate Pinning/CertificateHasher.swift
```diff
@@ -0,0 +1,129 @@
...
+        
+        return publicKeyHash as Data
+    }
+    
+    static func secKey(from certificate: SecCertificate) -> SecKey? {
+        var trust: SecTrust?
+        let status = SecTrustCreateWithCertificates(certificate, SecPolicyCreateBasicX509(), &trust)
+        
+        /*We do not need to evaluate trust here as we are simply looking to extract a public key - explicit evaluation is achieved by calling `checkValidity(of:)`.
+            This is useful for instance, when attempting to create a pinning hash for an invalid or expired certificate. If we were to validate on creation, we would not be able to test this. */
```
earlgaspard: Missing comma
```
This is useful, for instance, when
```

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+//  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+
+@available(iOSApplicationExtension 10.0, *)
+@available(tvOSApplicationExtension 10.0, *)
+@available(watchOSApplicationExtension 3.0, *)
+
+/* A configuration object which controls how a `TrustValidator` object makes decisions on which connections to accept and which to block. The configuration
```
earlgaspard: Can make this a multi-line documentation comment
```
/** A configuration object which controls
```

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+    /// - allow: Connections should be allowed after the certificate has expired, on the given `Date`.
+    /// - block: Connections should be blocked after the certificate has expired.
+    public enum CertificateExpirationPolicy: Hashable {
+        case allow(after: Date)
+        case block
+        
+        var expiration: Date? {
+            switch self {
+            case .allow(after: let date): return date
+            default: return nil
```
earlgaspard: can make switch exhaustive without `default`
```
switch self {
case .allow(after: let date): return date
case .block: return nil
```

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+        
+        /// Creates an instance of `DomainConfiguration` which will govern how any presented SSL certificate from that domain will be pinned.
+        ///
+        /// - Parameters:
+        ///   - domain: The domain for which this configuration is responsible. Should be entered in the format `example.com`.
+        ///   - enforced: When this value is set to true, a failure to pin a request will cause the connection to be blocked. Defaults to true.
+        ///   - includeSubdomains: Include any subdomains of the given domain in the pinning process. Defaults to true.
+        ///   - pinningHashes: An array of public key hashes that will be used to verify the presented SSL certificate.
+        ///   - expirationPolicy: The expiration policy to be used when creating this domain. A value of `.block` will cause an expired certificate to block all connections.
+        init(domain: String, enforced: Bool = true, includeSubdomains: Bool = true, pinningHashes: [Data], expirationPolicy: CertificateExpirationPolicy = .block) {
```
earlgaspard: Should this `init` be `public` also?

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+    
+    /// A list of `DomainConfiguration` objects which each represent a single domain's SSL pinning configuration.
+    public let domainConfigurations: [DomainConfiguration]
+    
+    // MARK: - Initializers
+    
+    /// Create a `TrustConfiguration` instance which will govern how the validator behaves when presented with a remote SSL certificate.
+    ///
+    /// - Parameter domainConfigurations: A list of `DomainConfiguration` objects which control the pinning process for each individual domain.
+    ///         There must only be a single domain configuration for a given domain. Multiple will trigger an assertion.
```
earlgaspard: This read better:
```
///         There must only be a single domain configuration for a given domain. Multiple domain configurations will trigger an assertion.
```

Sources/Certificate Pinning/TrustConfiguration.swift
```diff
@@ -0,0 +1,169 @@
...
+    public init(domainConfigurations: [DomainConfiguration]) {
+        assert(Set(domainConfigurations.map { $0.domain }).count == domainConfigurations.count, "You must not provided multiple domain configurations for any given domain.")
```
earlgaspard: Typo:
You must not provided multiple domain configurations -> You must not provide multiple domain configurations

Sources/Certificate Pinning/TrustValidatingNetworkService.swift
```diff
@@ -0,0 +1,80 @@
...
+    private let sessionTrustValidator: SessionDelegate
+    
+    // MARK: - Initializers
+    
+    /// Returns an instance of `TrustValidatingNetworkService` capable of performing server trust validation based on the provided configuration.
+    ///
+    /// - Parameters:
+    ///   - trustConfiguration: The configuration object used to govern the rules as to how server trust validation is performed on the network service.
+    ///   - sessionConfiguration: The configuration of the `URLSession` object used by the `NetworkService`.
+    ///   - networkActivityIndicatable: An object capable of displaying currently network activity visually to the user.
```
earlgaspard: Typo: currently -> current

Sources/Certificate Pinning/TrustValidator.swift
```diff
@@ -0,0 +1,102 @@
...
+//  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+
+@available(iOSApplicationExtension 10.0, *)
+@available(tvOSApplicationExtension 10.0, *)
+@available(watchOSApplicationExtension 3.0, *)
+
+/* This class performs validation on authentication challenges presented to it. In addition to ensuring that the challenge is trusted by the operating system, it will
```
earlgaspard: can use multi-line documentation comments here
```
/** This class
```

wmcginty:
@earlgaspard Changes have been made.

PR #85, opened 2019-02-19
Feature/cancellation

wmcginty:
OK, as promised here is a draft PR for what a revamp to the cancellation mechanism might look like. I have somewhat implemented a `CancellationSource` into the `NetworkService` and by extension the `BackendService`. 

1) When executing the request, keep and store a `CancellationSource`:

```swift
let source = CancellationSource()
preferredBackendService.execute(request: getUserRequest, cancellationToken: source.token) { [weak self] result in
            debugPrint("Get user result: \(result)")
            
            switch result {
            case .success(let user):
                self?.presentAlert(titled: "Fetched user", message: "\(user)")
            case .failure(let error):
                self?.presentAlert(titled: "Error", message: "\(error)")
      }
}
```

At this point we could definitely provide a customization point into the creation of the token if needed, for instance:

```swift
source.token(with: UUID())
``` 

Inside Hyperspace, the `BackendService` simply passes the `CancellationSource.Token` into the `NetworkService` where it is used in generation of the request:

```swift
public func execute(request: URLRequest, cancellationToken: CancellationSource.Token? = nil, completion: @escaping NetworkServiceCompletion) {
    let task = session.dataTask(with: request) { [weak self] (data, response, error) in
        //completion handler
    }
    
    cancellationToken?.register { task.cancel() }
    //more implementation
}
```

At this point, all you have to do from the client of the `BackendService` has to do is call:

```swift
source.cancel()
```

This will cancel all the tokens that have been generated by the given `source`. In this specific case, this results in a result of : `.failure(cancelled)`

### DIFF
--- a/Sources/Hyperspace/Service/Network/Cancellation.swift
+++ b/Sources/Hyperspace/Service/Network/Cancellation.swift
@@ -1,6 +1,6 @@
 //
 //  Cancellation.swift
-//  Hyperspace-iOS
+//  Hyperspace
 //
 //  Created by Will McGinty on 2/19/19.
 //  Copyright © 2019 Bottle Rocket Studios. All rights reserved.
--- a/Examples/Hyperspace-iOSExample/ViewController.swift
+++ b/Examples/Hyperspace-iOSExample/ViewController.swift
@@ -50,8 +50,8 @@ class ViewController: UIViewController {
 extension ViewController {
     private func getUser() {
         let getUserRequest = GetUserRequest(userId: 1)
-        
-        preferredBackendService.execute(request: getUserRequest) { [weak self] result in
+        let source = CancellationSource()
+        preferredBackendService.execute(request: getUserRequest, cancellationToken: source.token) { [weak self] result in
             debugPrint("Get user result: \(result)")
             
             switch result {
@@ -61,6 +61,8 @@ extension ViewController {
                 self?.presentAlert(titled: "Error", message: "\(error)")
             }
         }
+        
+        source.cancel()
     }
     
     private func createPost(titled title: String) {
--- a/Hyperspace.xcodeproj/project.pbxproj
+++ b/Hyperspace.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0E00FDD0221CB8BF00141E6A /* Cancellation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E00FDCF221CB8BF00141E6A /* Cancellation.swift */; };
 		0E29092D21349F600031F67C /* DecodingFailureTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E29092B21349F320031F67C /* DecodingFailureTests.swift */; };
 		0E29092E21349F610031F67C /* DecodingFailureTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E29092B21349F320031F67C /* DecodingFailureTests.swift */; };
 		0E347E5B2155A7A500BF18FA /* AnyError+Request.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E347E5A2155A7A500BF18FA /* AnyError+Request.swift */; };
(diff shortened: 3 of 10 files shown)


wmcginty:
@tylermilner @earlgaspard @GrandLarseny thoughts?

Sources/Hyperspace/Service/Network/Cancellation.swift
```diff
@@ -0,0 +1,78 @@
...
+    public init() {}
+    
+    // MARK: Interface
+    public var isCancelling: Bool {
+        lock.lock()
+        defer { lock.unlock() }
+        return observers == nil
+    }
+    
+    public var token: CancellationSource.Token { return CancellationSource.Token(source: self) }
```
earlgaspard: Isn't `CancellationSource` suppose to manage cancellation tokens? It only appears to keep track of one token.

Sources/Hyperspace/Service/Network/Cancellation.swift
```diff
@@ -0,0 +1,78 @@
...
+import Foundation
+
+/// Manages cancellation tokens and signals them when cancellation is requested.
+///
+/// All `CancellationTokenSource` methods are thread safe.
+public final class CancellationSource {
+    
+    public struct Token {
+        // MARK: Properties
+        private weak var source: CancellationSource?
```
earlgaspard: Seems strange to me that the token needs a reference to the source.

Sources/Hyperspace/Service/Network/Cancellation.swift
wmcginty: It keeps track of as many Token's as you create using it as a source. If you check out the `register` function of the `Token`, it calls into `CancellationSource.lockedRegister` which registers the closure as an observer (inserting into it's Array storage). As soon as you call `cancel` on the `CancellationSource` it runs it's entire array of observers to perform the cancellation.

Sources/Hyperspace/Service/Network/Cancellation.swift
wmcginty: This is so the closure used in the `register` method can be passed back to the `CancellationSource` so that synchronized cancellation can occur.

Sources/Hyperspace/Service/Network/Cancellation.swift
```diff
@@ -0,0 +1,78 @@
...
+        
+        // MARK: Interface
+        public func register(closure: @escaping () -> Void) {
+            source?.register(closure)
+        }
+    }
+    
+    // MARK: Properties
+    private let lock = NSLock()
+    private var observers: [() -> Void]? = []
```
tylermilner: What's the reasoning around this being an optional property that gets immediately set to an empty array?

wmcginty: They are used to represent different states - the initial state is an empty array to indicate their own observers. Upon cancellation we nil the array out and then use that `.none` state to indicate that cancellation of this source has already occurred, so any blocks that are registered from that point on will run immediately.

It's not super clear I agree, but a) this shouldn't really need to be 'understood' by the user b) it's going to improve performance a bit locking the single array vs locking the array and an `isCancelling` bool to keep state.

wmcginty:
Another thing I've been thinking about recently that _could_ work well with a system like this is priority. I've got a use case at the moment where the `URLSessionTask.Priority` of each of my requests can change drastically based on a user's scroll position (if it shouldn't be cancelled altogether). After a quick think, a `token` like system that represents the task in flight could work as well. 

Just a thought 🤷‍♂️

Sources/Hyperspace/Service/Network/Cancellation.swift
earlgaspard: So let me see if I understand this correctly:
I can create a `CancellationSource` and give it a `CancellationSource.Token`. I can register closures with the token which registers the closures in the source. When I ask the token for `isCancelling`, it asks the source for `isCancelling`. When I call `cancel()` on the source, it runs through the array of observers to perform the closures without concern for the token.
I am failing to see the need of having a token when all it does is forward or get things from the source.

wmcginty: You don't necessarily give the `CancellationSource` a `Token`. The source creates tokens on its own using the fileprivate initializer. For you, as the end user, to get a token, you just ask the source.

Let's say you have multiple concurrent tasks, or even a single task involving a series of serial subtasks. In most cases, you'll want to have 1 single source of cancellation. So in this case, if you tell the source to cancel, it will cancel each of its tokens automatically meaning all subtasks will be cancelled. If you only have 1 token active from a source, only 1 will be cancelled, but the API to do so remains the same whether your cancelling 1 or 20 tasks.

earlgaspard: That makes a bit more sense. So if I had several tokens and wanted to cancel just one of them, how would I do that? or is that beyond the scope of what this is intended to accomplish by being able to cancel all requests

wmcginty: We talked briefly about this and came up with some solutions

1) You could create your tokens each with different sources.
2) We could add additional identifying information (perhaps a UUID) to each token so that they can be differentiated from one another within a single cancellation source.

earlgaspard: My understanding is much improved. Thanks @wmcginty

PR #88, opened 2019-04-23
Swift5

wmcginty:
Initial migration of BrightFutures and Swift 5


Hyperspace.podspec
```diff
@@ -29,25 +29,18 @@ Unlike other networking libraries, the goal of Hyperspace is to keep things simp
   s.source           = { :git => 'https://github.com/BottleRocketStudios/iOS-Hyperspace.git', :tag => s.version.to_s }
   # s.social_media_url = 'https://twitter.com/<TWITTER_USERNAME>'
 
+  s.swift_version = '5'
```
tylermilner: I'm sure it doesn't matter, but would it be better to be explicit? `s.swift_version = '5.0'`

wmcginty: What an unreasonable request. Fixed.

PR #92, opened 2019-05-30
Add URL modification capability

wmcginty:


Tests/NetworkRequestTests.swift
```diff
@@ -153,6 +153,13 @@ class RequestTests: XCTestCase {
         XCTAssertNotNil(finalHeaders?[.authorization])
     }
     
+    func test_Request_ModifyingURL() {
+        var request = SimpleGETRequest()
+        
+        let final = request.usingURL(request.url.appendingQueryItems([URLQueryItem(name: "test", value: "value")]))
+        XCTAssertNotNil(final.url.query)
```
tylermilner: Would it be better to check the URL itself rather than just making sure a query was created?
```swift
XCTAssertEqual(final.url.absoluteString, "https://apple.com?test=value")
```

PR #95, opened 2019-07-13
Changed AnyError's implementation of NetworkServiceFailureInitializable

rickbdotcom:
To allow access to underlying failureResponse.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 #### Enhancements
 
-* None
-
+* Changed underlying error in AnyError's NetworkServiceFailureInitializable implementation from NetworkServiceError to NetworkServiceFailure so it can return its failure response rather than nil.
+[Richard Burgess](https://github.com/rickbdotcom)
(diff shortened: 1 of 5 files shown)


wmcginty:
Thanks for the contribution @rickbdotcom ! The changes look good to me - can you fix the resulting broken unit tests and add a changeling entry before merging?

rickbdotcom:
Will do!

wmcginty:
@rickbdotcom This looks great, if you can just add an entry to `Changelog.md` I think we're good to go!

rickbdotcom:
OK updated change log. Now one issue I can see with this change is that it could break some code if a consumer of the Framework was depending on the underlying error being NetworkServiceError and doing a cast of AnyError.error to NetworkServiceError.

wmcginty:
@rickbdotcom I don't think that's a huge issue assuming we bump the minor version number and call the impact of the change out. Thoughts @tylermilner ?

tylermilner:
@wmcginty sounds good to me. @rickbdotcom looks like there's a conflict in `CHANGELOG.md`. Could you resolve so that we can merge?

PR #101, opened 2019-07-30
Added Carthage Support

ganttastic:
Removed last remnants of Result framework. Removed BrightFutures project in the workspace as it's provided by carthage already built. Added carthage copy-frameworks scripts so it will pull in the correct carthage provided BrightFutures framework for each platform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![CI Status](https://img.shields.io/travis/BottleRocketStudios/iOS-Hyperspace/master.svg)](https://travis-ci.org/BottleRocketStudios/iOS-Hyperspace)
 [![Version](https://img.shields.io/cocoapods/v/Hyperspace.svg?style=flat)](http://cocoapods.org/pods/Hyperspace)
+[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -204,13 +205,26 @@ From here, you can open up `Hyperspace.xcworkspace` and run the examples:
 
 ## Installation
 
+### Cocoapods
```
tylermilner: Super minor, but let's add a blank line after this so that the formatting is consistent between the "Cocoapods" and "Carthage" sections.

tylermilner (dismissed):
Looks good. I was hoping this might fix the build warning for the tvOS target, but I guess it might be a problem with the BrightFutures framework itself. 🤷‍♂ 
image

wmcginty (dismissed):

README.md
ganttastic: Fixed!

ganttastic:
@wmcginty / @tylermilner - Between the two travis builds all platforms build and pass. 😄 Plus the previous build passed and I only updated the README. Is there a way to rebuild? Or is it good to go?

wmcginty:
There should be a way - if you go to the 'details' of the build - then the top right should have a 'restart build' button.

ganttastic:
@wmcginty / @tylermilner We're passing now. Good for review.

tylermilner (approved):

PR #102, opened 2019-08-05
Add SPM Support

ganttastic:
Testing Package.swift file with travis build. Using `swift build` to validate package.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -12,3 +12,4 @@ identifier_name:
 
 excluded:
   - Carthage
+  - .build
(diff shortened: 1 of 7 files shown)


ganttastic:
@tylermilner / @wmcginty - I've added Swift Package Manager support to Hyperspace. It works nicely in Xcode 11! I'm using the `swift build` command to test that the package is configured correctly. Similar to Carthage, there aren't really "subspecs". We could split it out into separate modules and targets, but I didn't see the need. Let me know what you guys think!

PR #105, opened 2019-09-16
WIP - Improve error details when the underlying request fails

tylermilner:
Added an associated type to the `unknownError` case of `NetworkServiceError`. Previously, the `Error` provided in the completion of the execution of a request was being thrown away if the `response` parameter was not a `HTTPURLResponse`. In some cases, these errors were internally mapped to a `NetworkServiceError` (`noInternetConnection`, `timedOut`, and `cancelled`), but now clients have access to all other errors that might occur (`NSURLErrorBadURL`, `NSURLErrorCannotFindHost`, `NSURLErrorDNSLookupFailed`, etc.).

Basically, instead of needing to manually map a lot more cases in [this switch](https://github.com/BottleRocketStudios/iOS-Hyperspace/blob/master/Sources/Hyperspace/Service/Network/NetworkServiceHelper.swift#L51), we now pass the underlying error through so that clients can have a more descriptive explanation of what went wrong vs just getting back a `NetworkServiceError.unknownError`.

### DIFF
--- a/Sources/Hyperspace/Service/Network/NetworkServiceProtocol.swift
+++ b/Sources/Hyperspace/Service/Network/NetworkServiceProtocol.swift
@@ -15,8 +15,8 @@
 import Foundation
 
 /// Represents an error that occurred when executing a Request using a NetworkService.
-public enum NetworkServiceError: Error, Equatable {
-    case unknownError
+public enum NetworkServiceError: Error {
+    case unknownError(Error?)
     case unknownStatusCode
     case redirection
     case clientError(HTTP.Status.ClientError)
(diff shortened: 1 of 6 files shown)


Sources/Hyperspace/Service/Network/NetworkServiceProtocol.swift
```diff
@@ -71,3 +71,33 @@ public protocol NetworkServiceProtocol {
...
 }
+
+// MARK: - Equatable Implementations
+
+extension NetworkServiceError: Equatable {
+    
+    public static func == (lhs: NetworkServiceError, rhs: NetworkServiceError) -> Bool {
+        switch (lhs, rhs) {
+        case (.unknownError(let lhsError), .unknownError(let rhsError)):
+            return lhsError?.localizedDescription == rhsError?.localizedDescription
```
tylermilner: This is the only piece of this that makes me a little uneasy. Unfortunately, I don't know of a better way to check for equality of two objects conforming to `Error`, but I think this should be safe for 99% of use cases. We were already doing this in our test code for `MockBackendServiceError`.

wmcginty: We could also just treat all `.unknownError` as equal, regardless of associated `Error`? Maybe?

tylermilner: We could, but I saw some recommendations against that on the [Swift forums](https://forums.swift.org/t/making-error-sub-enums-equatable/5878/6).

tylermilner:
I think #130 and version `4.0` in general should address a lot of these ideas. Closing...

PR #106, opened 2019-09-20
Add assertion to GET requests with body

wmcginty:


Sources/Hyperspace/Service/Backend/BackendService.swift
```diff
@@ -31,6 +31,8 @@ public class BackendService {
 
 extension BackendService: BackendServiceProtocol {
     public func execute<T: Request>(request: T, completion: @escaping BackendServiceCompletion<T.ResponseType, T.ErrorType>) {
+        assert(!(request.method == .get && request.body != nil), "An HTTP GET request should not contain request body data.")
+
```
rmirabelli: Would we prefer a precondition() here to really force the issue?

rmirabelli: I suppose it's a "must not" vs "should not" issue.

PR #110, opened 2019-12-18
Feature/recovery

wmcginty:


Sources/Futures/BackendService+Futures.swift
```diff
@@ -23,4 +24,19 @@ extension BackendServiceProtocol {
         
         return promise.future
     }
+
+    /// Executes the Request, returning a Future when finished.
+    ///
+    /// - Parameters:
+    ///   - recoverable: The Request to be executed, and is eligible to be recovered from in the case of failure.
```
tylermilner: Minor: The comment here refers to the parameter as `recoverable`, but it's named `request` in the method signature below.

wmcginty: Fixed - changed the signature so it's easy to delineate between the two when dynamic dispatch is involved.

tylermilner:
Codebeat is warning about block nesting being too deep. We can fix that in the future. I don't want to alter Codebeat settings right now and force this to go through another build/test cycle on Travis so forcing the merge right now.

Issue #120, opened 2020-06-09
Consider implementing `ExpressibleByStringLiteral` on HTTP header types


tylermilner:
This might make it a little nicer when you need to use customer header keys/values.
```swift
extension HTTP.HeaderValue: ExpressibleByStringLiteral {
    
    public init(stringLiteral value: StringLiteralType) {
        self.init(rawValue: value)
    }
}
```

This would allow you to quickly mix pre-defined and custom header values:
```swift
struct APIHeaders {
    static let defaultHeaders: [HTTP.HeaderKey: HTTP.HeaderValue] = [.authorization: "Basic abc123"]
}
```

Instead of needing to manually init the `HTTP.HeaderValue` every time:
```swift
struct APIHeaders {
    static let defaultHeaders: [HTTP.HeaderKey: HTTP.HeaderValue] = [.authorization: HTTP.HeaderValue(rawValue: "abc123")]
}
```

Issue #103, opened 2019-08-29
Add built-in support for x-www-form-urlencoded


tylermilner:
URL encoded forms are a pretty common mechanism of sending credentials to a backend for OAuth. We should add some built-in support for making it easy to generate a request in this format. Check out this code example on the Apple developer forums: https://forums.developer.apple.com/thread/113632.

PR #125, opened 2020-08-18
Feature/additional headers

wmcginty:


wmcginty:
Addresses #120  and #103

Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -100,31 +108,52 @@ public struct HTTP {
...
         }
         
-        /// Initializes a new `HTTP.Body` instance given an encodable object.
+        /// Returns a new `HTTP.Body` instance given an encodable object.
         /// - Parameters:
         ///   - encodable: The `Encodable` object to be included in the request.
         ///   - encoder: The `JSONEncoder` to be used to encode the object.
-        public init<E: Encodable>(_ encodable: E, encoder: JSONEncoder = JSONEncoder()) throws {
+        ///   - additionalHeaders: Any additional HTTP headers that should be sent with the request.
+        /// - Returns: Returns a new instance of `HTTP.Body` with the given encodable representation.
```
earlgaspard, line 128: I think the second `Returns` can be removed.
```
/// - Returns: A new instance of `HTTP.Body` with the given encodable representation.
```

Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -100,31 +108,52 @@ public struct HTTP {
...
+        /// Returns a new `HTTP.Body` instance given an encodable object.
         /// - Parameters:
         ///   - encodable: The `Encodable` object to be included in the request.
         ///   - container: A type of `EncodableContainer` in which to encode the object.
         ///   - encoder: The `JSONEncoder` to be used to encode the object.
-        public init<E, C: EncodableContainer>(_ encodable: E, container: C.Type, encoder: JSONEncoder = JSONEncoder()) throws where C.Contained == E {
-            let data = try encoder.encode(encodable, in: container)
-            self.init(data)
+        ///   - additionalHeaders: Any additional HTTP headers that should be sent with the request.
+        /// - Returns: Returns a new instance of `HTTP.Body` with the given encodable representation.
```
earlgaspard, line 141: Same here about removing the second `Returns`.

Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -100,31 +108,52 @@ public struct HTTP {
...
+            throws -> HTTP.Body where C.Contained == E {
+                let data = try encoder.encode(encodable, in: container)
+                return HTTP.Body(data, additionalHeaders: additionalHeaders)
+        }
+
+        /// Initializes a new `HTTP.Body` instance given a set of URL form content
+        /// - Parameters:
+        ///   - formContent: An array of `(String, String)` representing the content to be encoded.
+        ///   - additionalHeaders: Any additional HTTP headers that should be sent with the request.
+        /// - Returns: Returns a new instance of `HTTP.Body` with the given form content.
```
earlgaspard, line 153: Same here about removing the second `Returns`.

Sources/Hyperspace/HTTP/HTTP.swift
wmcginty: Fixed

Sources/Hyperspace/HTTP/HTTP.swift
```diff
@@ -100,31 +108,52 @@ public struct HTTP {
     public struct Body: Equatable {
         
         /// The raw body data to be attached to the HTTP request
-        public let data: Data
+        public let data: Data?
+        public let additionalHeaders: [HeaderKey: HeaderValue]
```
tylermilner, line 112: Seeing how this is used, I'm wondering if we should abstract the concept up to a `contentType` property? I suppose this approach is technically more flexible, but I would imagine you would usually want to assign as many of your headers directly on the `Request` itself. Thoughts?
```swift
public let contentType: HeaderValue?
```

wmcginty: That could work as well, for sure. I'm trying to think of any times I've wanted to add 2 header key/value pairs when I've set a particular HTTP body and coming up empty (not sure if any come to anyone else's mind).

It could also be another convenience initializer?

```swift
init(data: Data?, contentType: HeaderValue?) {
 self.init(data: data, additionalHeaders: contentType.map { [.contentType: $0] })
}
```

tylermilner: That could work, though now that I'm thinking about it, we might already have what we need with the static `json(_:encoder:additionalHeaders:)` and `urlForm(using:additionalHeaders:)` functions that you created below. I think we'll pretty much always be using those when we need to construct a `HTTP.Body`, right?

wmcginty: Yea, the main reason I only left it defaulted is there are some services that have some weirdness associated with it -
for instance Airship uses a custom `application/vnd.urbanairship+json; version=3` content type for JSON. I can't personally think of anywhere that would need 2 headers, but I'm almost positive there's someone out there 😄

PR #132, opened 2020-12-24
Decodable EmptyResponse

BR-Nathan-Chiu:
Making EmptyResponse conform to Decodable to avoid branching logic between empty responses and JSON responses


wmcginty:
I believe the reason this is not marked as `Decodable` is because many projects have different requirements for an empty response. Some might mean 'any response' while other's might need a completely empty response. This seems like it would break that.

If you are using master, does `EmptyDecodingStrategy` solve your problem? There is a `.default` strategy which will take any response as a success, and a `.validatedEmpty` strategy which will ensure the response is empty before succeeding. More info here: https://github.com/BottleRocketStudios/iOS-Hyperspace/blob/master/Documentation/Guides/Handling%20Empty%20Responses.md (It's still in 'beta', so any feedback is appreciated)

PR #142, opened 2021-10-31
Feature/async await

GrandLarseny:
Y'all do these comments long, but I'm not so sure about that. Anyway, here's what I did:

- Added in an extension to `BackendServiceProtocol` to shim in an `async` function.
- Added unit tests for said `async` shim.
- Updated Xcode project to work with Xcode 12 and later.
- Updated Carthage to get project to compile (per instructions in `README.md`)
- Updated platform target for watchOS to watchOS 6 (necessary to get project compiling)

Some things I'd like to do either in this PR or in a future PR:

- Get sample `async` calls in playground and example project
- Rework project structure around SPM authoring
- Make `async` calls a bit more native to the framework once supported by the Foundation framework. That means removing the `async withCheckedThrowingContinuation` block and adding in `async` versions of the execute function to the transport and backend layers.

### DIFF
--- a/Sources/Async/BackendService+Async.swift
+++ b/Sources/Async/BackendService+Async.swift
@@ -1,0 +1,29 @@
+//
+//  BackendService+Async.swift
+//  Hyperspace-iOS
+//
+//  Created by Daniel Larsen on 10/30/21.
+//  Copyright © 2021 Bottle Rocket Studios. All rights reserved.
+//
+
+import Foundation
+
+extension BackendServiceProtocol {
+
+    /// Executes the Request, returning the specified value type. Throws an error if unable to return the value as specified.
+    ///
+    /// - Parameters:
+    ///   - request: The Request to be executed.
+    /// - Returns: The decodable type specified in the `Request`.
+    @available(iOSApplicationExtension 13.0.0, *)
+    public func execute<T, U>(request: Request<T, U>) async throws -> T {
+        return try await withCheckedThrowingContinuation { continuation in
+            execute(request: request) { result in
+                switch result {
+                case .success(let value): continuation.resume(returning: value)
+                case .failure(let err): continuation.resume(throwing: err)
+                }
+            }
+        }
+    }
+}
--- a/Sources/Async/BackendService+Async.swift
+++ b/Sources/Async/BackendService+Async.swift
@@ -8,20 +8,36 @@
 
 import Foundation
 
+@available(iOS 13.0, *)
 extension BackendServiceProtocol {
 
-    /// Executes the Request, returning the specified value type. Throws an error if unable to return the value as specified.
+    /// Executes the Request asynchronously, returning the specified value type. Throws an error if unable to return the value as specified.
     ///
     /// - Parameters:
     ///   - request: The Request to be executed.
     /// - Returns: The decodable type specified in the `Request`.
-    @available(iOSApplicationExtension 13.0.0, *)
     public func execute<T, U>(request: Request<T, U>) async throws -> T {
         return try await withCheckedThrowingContinuation { continuation in
             execute(request: request) { result in
                 switch result {
                 case .success(let value): continuation.resume(returning: value)
                 case .failure(let err): continuation.resume(throwing: err)
+                }
+            }
+        }
+    }
+
+    /// Executes the Request asynchronously without throwing any errors
+    ///
+    /// - Parameters:
+    ///   - request: The Request to be executed.
+    /// - Returns: A `Result` of given value type `T` and error type `U`.
+    public func executeWithResult<T, U>(request: Request<T, U>) async -> Result<T, U> {
+        return await withCheckedContinuation { continuation in
+            execute(request: request) { result in
+                switch result {
+                case .success(let value): continuation.resume(with: .success(Result<T, U>.success(value)))
+                case .failure(let err): continuation.resume(with: .success(Result<T, U>.failure(err)))
                 }
             }
         }
--- a/Sources/Async/BackendService+Async.swift
+++ b/Sources/Async/BackendService+Async.swift
@@ -17,13 +17,11 @@ extension BackendServiceProtocol {
     ///   - request: The Request to be executed.
     /// - Returns: The decodable type specified in the `Request`.
     public func execute<T, U>(request: Request<T, U>) async throws -> T {
-        return try await withCheckedThrowingContinuation { continuation in
-            execute(request: request) { result in
-                switch result {
-                case .success(let value): continuation.resume(returning: value)
-                case .failure(let err): continuation.resume(throwing: err)
-                }
-            }
+        let result = await executeWithResult(request: request)
+
+        switch result {
+        case .success(let value): return value
+        case .failure(let err): throw err
         }
     }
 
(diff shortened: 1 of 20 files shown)


Sources/Async/BackendService+Async.swift
```diff
@@ -0,0 +1,29 @@
...
+
+extension BackendServiceProtocol {
+
+    /// Executes the Request, returning the specified value type. Throws an error if unable to return the value as specified.
+    ///
+    /// - Parameters:
+    ///   - request: The Request to be executed.
+    /// - Returns: The decodable type specified in the `Request`.
+    @available(iOSApplicationExtension 13.0.0, *)
+    public func execute<T, U>(request: Request<T, U>) async throws -> T {
```
wmcginty, line 19: I would also like to see a non-throwing version of this that doesn't lose `U`'s type information. Something like 

```public func execute<T, U>(request: Request<T, U>) async -> Result<T, U>```

Doing it only this way, it feels like having the error type strongly typed on `Request` is useless, because you'll have to catch all errors anyway.

**Review by wmcginty — COMMENTED** [review]
This is looking good @GrandLarseny. I also agree that centering the project structure around SPM going forward is a good idea, but probably for a different PR.

I left a comment about a(n additional) version of `execute` that preserves `Error` type information, curious to hear other people's thoughts.

File: Sources/Async/BackendService+Async.swift (hunk 1 above)
Comment (GrandLarseny) [reply Sources/Async/BackendService+Async.swift:19]: Thanks so much for the feedback! I definitely agree that having strongly typed errors in the result is good, and being able to distinguish errors by type is useful.

Here's what I'm thinking, though. Doing a generic `do ... catch` could be a bit more true to the process. There are many different types of errors that could occur during the execution of the request in Hyperspace. Wouldn't it make sense to be able to throw all those errors as the type they are, not necessarily wrapped in another type?

But then we get to your question of why even specify the error type in the `Request`. And I agree that it becomes less useful, but it still might have a use when the response is an error code, but contains a valid body that can be decoded. 


To be completely honest, I feel like I'm missing something, though. Do you have any specific examples I can use to make sure that your use case is being covered?

Comment (wmcginty) [reply Sources/Async/BackendService+Async.swift:19]: > Here's what I'm thinking, though. Doing a generic do ... catch could be a bit more true to the process. There are many different types of errors that could occur during the execution of the request in Hyperspace. Wouldn't it make sense to be able to throw all those errors as the type they are, not necessarily wrapped in another type?

I disagree- if I create a `Request<X, MyError>`, then I would expect to get back either `X` or an instance of `MyError`. Conformance of `MyError` to `TransportFailureRepresentable` is needed to ensure any transport failure can be converted to `MyError`, and optionally, conformance to `DecodingFailureRepresentable` is used to automatically handle JSON decoding errors.

In my opinion, the entire benefit of those two protocols is to encapsulate and surface the multitude of things that can go wrong in a single, easy-to-reason-with, type. If we think it would be better to throw all those errors as the type they are, then we could just extend `URLSession` to automatically try to decode something.

I'm not saying that something like `public func execute<T, U>(request: Request<T, U>) async -> Result<T, U>`, should be the only option (I think the majority of people who use `AnyError` would generally default to using a general `async throws` as they aren't getting much specific use out of `AnyError` anyway), but I do think it should be an option.

For example, let's say I have a type `MyError`. It conforms to `TransportFailureRepresentable`, so any time I execute a closure-based request, the compiler is guaranteeing the type of any error I receive. 

```swift
switch result {
case .success: break
case .failure(let error): //error is MyError
```

Using the `async` version though (basically because `throws` is currently untyped) I lose that guarantee, and have to handle a case that shouldn't be possible based on how `BackendService` is implemented.

```swift
let request<X, MyError> = ...
do {
 let myX = try await backendService.execute(request)
catch let error as MyError {
  // Handle my error
} catch {
  // Handle a case that shouldn't be possible, but I have to handle each time anyway
}
```

GrandLarseny: Thanks so much for your thorough explanation, @wmcginty ! Given the use case you're laying out I do agree that a second function returning `Result<T, U>` does make the most sense. I'll add that and add unit tests for it as well.

I do think it'd be interesting to catch up on strategic approaches sometime. I still lean towards using the native `do ... catch` infrastructure over switching on a result, but I'd love to hear your experiences!

In the immediate term, though, I will add the async result function.

wmcginty: Thanks@GrandLarseny! I see that approach as well, and I think when Hyperspace is fully built on `async` functions, the entire library might look a bit different, more geared to `do`, `try`, `catch`.

Sources/Async/BackendService+Async.swift
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+    }
+
+    /// Executes the Request asynchronously without throwing any errors
+    ///
+    /// - Parameters:
+    ///   - request: The Request to be executed.
+    /// - Returns: A `Result` of given value type `T` and error type `U`.
+    public func executeWithResult<T, U>(request: Request<T, U>) async -> Result<T, U> {
```
GrandLarseny, line 35: I ended up giving preference to `public func execute<T, U>(request: Request<T, U>) async throws -> T` over this function when naming. Since both functions have the same signatures, renaming one is required for disambiguation. 

To be honest, I was surprised that it wasn't enough to disambiguate with `throw`ing vs non-throwing. But the compiler was complaining. 

I'm open to any suggestions for better naming!

wmcginty: Ah yea, makes sense - the compiler can't tell you whether you want a `T` or `Result<T, U>` unless you were to explicitly type the output. I'm ok with the difference in names, to be honest (at the very least, I can't think of anything better!)

Sources/Async/BackendService+Async.swift
```diff
@@ -0,0 +1,45 @@
...
+        return await withCheckedContinuation { continuation in
+            execute(request: request) { result in
```
wmcginty, line 37: No strong opinion here, but would it be worth implementing 1 of these functions using the other so there's only one explicit use of `withCheckedContinuation` ?

wmcginty (changes requested):
This looks great @GrandLarseny, the only thing I think that's needed at this point is a Changelog entry!

Sources/Async/BackendService+Async.swift
GrandLarseny: Sounds good, I'll make those changes

wmcginty (approved):

Sources/Certificate Pinning/CertificateHasher.swift
```diff
@@ -53,10 +53,10 @@ struct CertificateHasher {
     }
     
     static func checkValidity(of trust: SecTrust) -> Bool {
-        var result = SecTrustResultType.invalid
-        let status = SecTrustEvaluate(trust, &result)
+        var result: CFError?
+        let status = SecTrustEvaluateWithError(trust, &result)
         
-        return (status == errSecSuccess) ? (result == .unspecified || result == .proceed) : false
+        return status || result == nil
```
br-tyler-milner, line 59: I realize I'm a bit late to the party, but is there a reason why we adjusted the certificate pinning logic in this PR?

GrandLarseny: `SecTrustEvalute()` has been deprecated in favor of `SecTrustEvaluateWithError(_:_:) `

PR #143, opened 2021-11-10
Feature/async await refactor

GrandLarseny:
I was having problems using the async functions in a project. Seems like it might be a namespace collision with the futures-based execution function.

### DIFF
--- a/Sources/Async/BackendService+Async.swift
+++ b/Sources/Async/BackendService+Async.swift
@@ -16,7 +16,7 @@ extension BackendServiceProtocol {
     /// - Parameters:
     ///   - request: The Request to be executed.
     /// - Returns: The decodable type specified in the `Request`.
-    public func execute<T, U>(request: Request<T, U>) async throws -> T {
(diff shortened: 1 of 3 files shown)


Sources/Async/BackendService+Async.swift
```diff
@@ -16,7 +16,7 @@ extension BackendServiceProtocol {
     /// - Parameters:
     ///   - request: The Request to be executed.
     /// - Returns: The decodable type specified in the `Request`.
-    public func execute<T, U>(request: Request<T, U>) async throws -> T {
+    public func executeAsync<T, U>(request: Request<T, U>) async throws -> T {
```
wmcginty, line 19: Should we rename `executeWithResult` similarly (even though it doesn't have a similar naming issue currently) ?

GrandLarseny: Ugh, nevermind. Gonna delete this PR because I'm a goober who tried to implement the `async execute` wrong.

There's no namespace problem, it was an error I should have encountered earlier with an improperly defined `Request`. 

It's all working now.

PR #145, opened 2021-11-30
Wrap access to tasks variable in dispatch queue async

GrandLarseny:
Wrap references to `TransportService` variable `tasks` in a serialized dispatch queue.

### DIFF
--- a/Sources/Hyperspace/Service/Transport/TransportService.swift
+++ b/Sources/Hyperspace/Service/Transport/TransportService.swift
@@ -14,7 +14,7 @@ public class TransportService {
     // MARK: - Properties
     
     let session: TransportSession
-    private let queue = DispatchQueue(label: "Hyperspace.tasks.sync")
+    private let queue = DispatchQueue(label: "com.bottlerocketstudios.hyperspace.tasks")
     private(set) var networkActivityController: NetworkActivityController?
     private var tasks = [URLRequest: TransportDataTask]()
     
@@ -47,19 +47,19 @@ extension TransportService: Transporting {
         queue.async { [weak self] in
             self?.tasks[request] = task
             task.resume()
-
-            self?.networkActivityController?.start()
         }
+
+        networkActivityController?.start()
     }
     
     public func cancelTask(for request: URLRequest) {
-        queue.async { [weak self] in
+        queue.sync { [weak self] in
             self?.tasks[request]?.cancel()
         }
     }
     
     public func cancelAllTasks() {
-        queue.async { [weak self] in
+        queue.sync { [weak self] in
             self?.tasks.forEach { $1.cancel() }
         }
     }
--- a/Sources/Hyperspace/Service/Transport/TransportService.swift
+++ b/Sources/Hyperspace/Service/Transport/TransportService.swift
@@ -14,6 +14,7 @@ public class TransportService {
     // MARK: - Properties
     
     let session: TransportSession
+    private let queue = DispatchQueue(label: "Hyperspace.tasks.sync")
     private(set) var networkActivityController: NetworkActivityController?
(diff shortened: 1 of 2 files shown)


Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -42,19 +43,25 @@ extension TransportService: Transporting {
...
-        tasks[request] = task
-        task.resume()
-        
-        networkActivityController?.start()
+
+        queue.async { [weak self] in
+            self?.tasks[request] = task
+            task.resume()
+
+            self?.networkActivityController?.start()
```
wmcginty, line 51: In the interest of symmetry, we should probably either serially `start` and `stop`, or neither. If I remember correctly, `NetworkActivityController` already has an internal queue, so I'd recommend we leave it out of this `queue` altogether.

Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -42,19 +43,25 @@ extension TransportService: Transporting {
...
+        }
     }
     
     public func cancelTask(for request: URLRequest) {
-        tasks[request]?.cancel()
+        queue.async { [weak self] in
```
wmcginty, line 56: I feel like this should break some tests no? I think, because this is a `private` queue we're using, and we're careful how we use it we can safely `queue.sync` as well, which might fix any of the broken assumptions in the tests.

wmcginty (changes requested):

Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -14,6 +14,7 @@ public class TransportService {
     // MARK: - Properties
     
     let session: TransportSession
+    private let queue = DispatchQueue(label: "Hyperspace.tasks.sync")
```
wmcginty, line 17: Went back and checked and `NetworkActivityController` does have it's own queue, so regardless of the solutions for the below, I think we should match it's label `com.bottlerocketstudios.hyperspace.tasks` (or something).

wmcginty:
Also, in addition to 4.0.0 it may be worth back porting this into to a `3.2.2` release afterwards. There have been a lot of changes since then, but this might be worth applying to the existing latest release. Thoughts? @GrandLarseny  @BottleRocketStudios/team-ios-open-source-w

Sources/Hyperspace/Service/Transport/TransportService.swift
GrandLarseny: Fair point, I'll move it out of the queue.

GrandLarseny:
Sure, I can back port this to 3.2.2. Shouldn't take too long.

Sources/Hyperspace/Service/Transport/TransportService.swift
GrandLarseny: Hah, this gave me some pause, but ultimately I do think we can make the call synchronous.

Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -42,19 +43,25 @@ extension TransportService: Transporting {
             self?.networkActivityController?.stop()
             self?.handle(data: data, response: response, error: error, for: HTTP.Request(urlRequest: request), completion: completion)
         }
-        
-        tasks[request] = task
-        task.resume()
-        
+
+        queue.async { [weak self] in
```
wmcginty, line 47: I have less strong feelings on this one to be honest, but is this intentionally `async` ?

wmcginty: Works for me

wmcginty (approved):

PR #148, opened 2021-12-02
Feature/GitHub actions

wmcginty:
- Remove `BrightFutures` dependency
- Add GitHub Actions CI support
- Fixed a few issues with `async` / `await` and tvOS and watchOS


Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -44,7 +44,7 @@ extension TransportService: Transporting {
             self?.handle(data: data, response: response, error: error, for: HTTP.Request(urlRequest: request), completion: completion)
         }
 
-        queue.async { [weak self] in
+        queue.sync { [weak self] in
```
wmcginty, line 47: This was breaking a test, so in the interest of not making sweeping changes in what is essentially a CI PR, I made this asynchronous call synchronous. @GrandLarseny .

Sources/Hyperspace/Service/Transport/TransportService.swift
```diff
@@ -30,7 +30,7 @@ public class TransportService {
     }
     
     deinit {
-        cancelAllTasks()
+        tasks.forEach { $1.cancel() }
```
wmcginty, line 33: It looks like there is an "issue" with the synchronous access in `deinit`. `self` is available before we synchronously enter the queue, but it is nil when we are in the queue, meaning the tasks aren't cancelled.  In the interest of not making sweeping changes in what is essentially a CI PR, I made this asynchronous call synchronous. @GrandLarseny, definitely worth looking at before we ship 4.0.0

GrandLarseny (approved):
Hey, unit tests are working for me on this branch! Hooray! Great work!

Anyway, I see the problem you're solving here and I agree that using a synchronous block makes sense here.

wmcginty:
Yea, I think the `deinit` functionality could use some brainstorming on possible alternates. It feels kinda weird to just pull the functionality from `cancelAllTasks()` into the deinit.

Maybe
```swift
func cancelAllTasks(executeImmediately: Bool = false) {
  // use queue or not depending on executeImmediately
}

Thoughts @GrandLarseny ?

wmcginty:
@GrandLarseny  Updated to use Xcode 13.2, so I dropped the availability for the `async` functions back down to 13.0.

br-earl-gaspard (approved):

PR #152, opened 2022-02-03
Feature/add mac os example

andrew-winn-br:
Had to add Catalyst support because a Catalyst app uses the iOS/Catalyst version of a framework, not the macOS version.
This should complete #6 
Screen Shot 2022-02-03 at 11 54 36 AM
Screen Shot 2022-02-03 at 11 54 17 AM


andrew-winn-br:
Linter error is:

andrew-winn-br:
Linter error is:
```
PackageDescription.SupportedPlatform:5:24: note: 'macCatalyst' was introduced in PackageDescription 5.5
PackageDescription.SupportedPlatform:33:27: note: 'v13' was introduced in PackageDescription 5.5
```
and the current package description for Hyperspace is 5.1. I think I can just remove the .`macCatalyst(.v13)` line from Package.Swift and it should still be fine. `.macCatalyst` was added as a specific platform for SPM in Xcode 13, but Catalyst was supported before that.

wmcginty:
@andrew-winn-br  I think we go ahead and upgrade the minimum tools version to 5.5, especially as this isn't going straight into a release we don't have to immediately worry about compatibility issues.

Example/Sources/macOS/ViewController.swift
```diff
@@ -0,0 +1,96 @@
+//
+//  ViewController.swift
+//  macOS Example
+//
+//  Created by Andrew Winn on 2/2/22.
```
wmcginty, line 5: Can you remove the `Created by` headers? Just so know one tries to track you down specifically months from now when the correct method would be creating an issue on the project.

wmcginty (changes requested):
This mostly looks good to me, but can you also add a CHANGELOG.md entry?

Example/Sources/macOS/ViewController.swift
andrew-winn-br: Oops Yeah will do

andrew-winn-br: done

wmcginty (approved):

Example/Sources/macOS/ViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+                self?.presentAlert(titled: "Error", message: "\(error)")
+            }
+        }
+    }
+
+    private func deletePost(postId: Int) {
+        let deletePostRequest = Request<EmptyResponse, AnyError>.deletePost(withID: postId)
+
+        backendService.execute(request: deletePostRequest) { [weak self] result in
+            switch result {
```
br-earl-gaspard, line 71: I noticed `debugPrint` statements for other `backendService.execute` calls but not one here.

br-earl-gaspard (changes requested):

br-earl-gaspard (approved):

Example/Sources/macOS/ViewController.swift
andrew-winn-br: Done!

wmcginty (approved):

PR #155, opened 2022-04-19
Properly encode ampersand in forms

andrew-winn-br:


Sources/Core/Request/FormURLEncoder.swift
```diff
@@ -44,7 +44,7 @@ extension CharacterSet {
     static let urlFormAllowed: CharacterSet = {
         var allowed = CharacterSet.urlQueryAllowed
         allowed.insert(" ")
-        allowed.remove(charactersIn: "+/?")
+        allowed.remove(charactersIn: "+/?&")
```
wmcginty, line 47: I think we should change the `CharacterSet` to be 
```swift
CharacterSet(charactersIn: "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-._* ")
```

So that we're more explicit by what's allowed - reference [here](https://url.spec.whatwg.org/#urlencoded-serializing)

wmcginty (commented):
One tweak to the definition of the character set, otherwise this looks good!

Sources/Core/Request/FormURLEncoder.swift
wmcginty: Note the space isn't actually allowed, but it already has the special handling to replace with "+" above.

andrew-winn-br: Updated

wmcginty (approved):

br-earl-gaspard (approved):

PR #167, opened 2022-12-21
Add pretty printing to `HTTP.Response`

wmcginty:


br-earl-gaspard (dismissed):

br-earl-gaspard (approved):

Sources/Core/HTTP/HTTP.swift
```diff
@@ -224,6 +224,19 @@ public struct HTTP {
...
             return body.flatMap { String(data: $0, encoding: .utf8) }
         }
+
+        /// Returns a string of formatted JSON.
+        ///
+        /// - Parameters:
+        ///     - options: Options for creating the JSON data. Default is `prettyPrinted`.
+        func formattedJSON(with options: JSONSerialization.WritingOptions = [.prettyPrinted]) -> String? {
+            guard let data = body,
+                  let object = try? JSONSerialization.jsonObject(with: data, options: []),
```
rmirabelli, line 234: Can we do something on errors? Right here we're just swallowing problems with serialization, rather than passing the error up the chain.

We could even just make the function `throw` the errors upwards.

wmcginty: Yea, I can definitely make this a `throw`ing function instead.

rmirabelli: Awesome!

rmirabelli (approved):

br-earl-gaspard (approved):

PR #171, opened 2023-07-26
Build out PreparationStrategy functionality

wmcginty:
Adds a property on `BackendService` for an array of `PreparationStrategy`. These are executed in order just before every request is sent, allowing for just-in-time modifications to the request before send. They execute `async`, and can `throw`. Errors thrown from a `PreparationStrategy` are not recoverable using the existing`RecoveryStrategy` machinery and throw to the caller immediately (they do not hit the server). This is useful for applying a dynamic property to every request that isn't as simple as setting a common header.

For example
```swift
struct SignatureApplyingPreparationStrategy: PreparationStrategy {

    func prepare<R>(toExecute request: Request<R>) async throws -> Request<R> {
        return request.addingHeaders([.signature: .init(rawValue: applySignature(using: request.url, body: request.body))])
    }

    func applySignature(using url: URL, body: Data?) -> String {
        // ...
    }
}

let backendService = BackendService(preparationStrategies: SignatureApplyingPreparationStrategy())
try await backendService.execute(request: myRequest)
```

Whereas previously this signature would need to be applied to each request, the `PreparationStrategy` allows this logic to reside in one place, apply to all outgoing requests, but still be based on the properties of each individual request.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,6 +5,12 @@ disabled_rules:
   - trailing_whitespace # Disables SwiftLint complaining about whitespace characters on empty lines
   - todo # Disables auto-warning of TODO statements
 
+nesting:
+  type_level:
+    warning: 2
+  function_level:
+    warning: 3
+
 identifier_name:
   excluded:
     - id
--- a/Sources/Service/Peripheral/PreparationStrategy.swift
+++ b/Sources/Service/Peripheral/PreparationStrategy.swift
@@ -14,6 +14,6 @@ public protocol PreparationStrategy {
     /// - Parameter request: The request that is about to be executed
     /// - Returns: A modified version of the `Request` that will be executed.
     ///
-    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any of the`RecoveryStrategy` attached to the executing `BackendService`.
+    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any of the `RecoveryStrategy` attached to the executing `BackendService`.
     func prepare<R>(toExecute request: Request<R>) async throws -> Request<R>
 }
--- a/Sources/Service/Peripheral/PreparationStrategy.swift
+++ b/Sources/Service/Peripheral/PreparationStrategy.swift
@@ -14,6 +14,6 @@ public protocol PreparationStrategy {
     /// - Parameter request: The request that is about to be executed
     /// - Returns: A modified version of the `Request` that will be executed.
     ///
-    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any  of the`RecoveryStrategy` attached to the executing `BackendService`.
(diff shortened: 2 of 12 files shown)


rmirabelli (dismissed):

.swiftlint.yml
```diff
@@ -5,6 +5,12 @@ disabled_rules:
   - trailing_whitespace # Disables SwiftLint complaining about whitespace characters on empty lines
   - todo # Disables auto-warning of TODO statements
 
+nesting:
```
br-tyler-milner, line 8: Now that we have this `nesting` exception in `.swiftlint.yml`, we can remove the `swiftlint:disable` and `swiftlint:enable` comments out of `HTTP.swift`.

Sources/Service/Peripheral/PreparationStrategy.swift
```diff
@@ -0,0 +1,19 @@
...
+
+import Foundation
+
+public protocol PreparationStrategy {
+
+    /// Handle any just-in-time transformations needed on the request before execution.
+    /// - Parameter request: The request that is about to be executed
+    /// - Returns: A modified version of the `Request` that will be executed.
+    ///
+    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any  of the`RecoveryStrategy` attached to the executing `BackendService`.
```
br-tyler-milner, line 17: Small double-space typo.
```suggestion
    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any of the`RecoveryStrategy` attached to the executing `BackendService`.
```

br-tyler-milner (changes requested):

Sources/Service/Peripheral/PreparationStrategy.swift
wmcginty: Eagle eye - fixed.

.swiftlint.yml
wmcginty: Fixed

Sources/Service/Peripheral/PreparationStrategy.swift
```diff
@@ -0,0 +1,19 @@
...
+
+import Foundation
+
+public protocol PreparationStrategy {
+
+    /// Handle any just-in-time transformations needed on the request before execution.
+    /// - Parameter request: The request that is about to be executed
+    /// - Returns: A modified version of the `Request` that will be executed.
+    ///
+    ///  While this method is both `async` and `throws`, any errors thrown as part of the preparation are not recoverable using any of the`RecoveryStrategy` attached to the executing `BackendService`.
```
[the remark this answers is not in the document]
wmcginty: Fixed. English is hard :(

br-tyler-milner (approved):

wmcginty:
@rmirabelli Do you mind taking a look at this when you get a chance?

rmirabelli (approved):

rmirabelli:
Sorry I didn't have much to say, I think I looked at the preparation strategies more broadly a week or two ago.

